PR #1, opened 2021-03-08
Extract Iovec library

CraigFe:
Changes:
- split source into `lib/iovec` and `lib/uring`;
- namespace internal concepts a little more;
- vendor inside `vendor/` dir.

Follow-up PR coming that adds more bindings needed for `mirage/index` IO.


talex5:
Looks reasonable to me.

I'm just trying to enable CI for this repository, but the jobs are failing with `ENOMEM`. Strangely, if I get a shell in the running container and run the tests from there then it works!

talex5:
OK, CI is fixed now. I rebased your branch... I think the remaining errors are actual problems :-)

lib/iovec/iovec_stubs.c
```diff
@@ -0,0 +1,47 @@
...
+    iovs[i].iov_base = Caml_ba_data_val(v_ba);
+    iovs[i].iov_len = Caml_ba_array_val(v_ba)->dim[0];
+    dprintf( "iov %d: %p %lu\n", i, iovs[i].iov_base, iovs[i].iov_len);
+  }
+  if (((uintptr_t)iovs & 1) == 1)
+    caml_failwith("unaligned alloc??");
+  CAMLreturn((value)iovs | 1);
+}
+
+value ocaml_iovec_advance(value v_iovecs, value v_idx, value v_adj)
```
avsm, line 31: This horrendously unsafe function does need to go in a future iteration ;-)

PR #8, opened 2021-04-14
Add support for `noop` command

CraigFe:
... and refactor some of the tests to use Alcotest. Currently requires the dev version of Alcotest, which I will release shortly.


lib/uring/uring.ml
```diff
@@ -65,9 +66,12 @@ type 'a t = {
   mutable dirty: bool; (* has outstanding requests that need to be submitted *)
 }
 
+let invalid_argf fmt = Format.kasprintf invalid_arg fmt
```
talex5, line 69: We already depend on `fmt`, so might as well use `Fmt.invalid_arg` instead.

tests/main.ml
```diff
@@ -0,0 +1,86 @@
...
+
+module Test_data = struct
+  let path = "output_file.txt"
+
+  let setup () =
+    let oc = open_out path in
+    output_string oc "A test file";
+    close_out oc
+
+  let get_fd () = Unix.openfile path [ O_RDONLY ] 0
```
talex5, line 13: Maybe `with_fd`, to avoid leaking it?

talex5 (commented):
Thanks for cleaning this up!

tests/main.ml
CraigFe: Good idea; done.

CraigFe:
The CI is now happy with this. I'll make some follow-up PRs shortly.

PR #9, opened 2021-04-22
Re-implement the user_data heap

CraigFe:
In benchmarking Uring + Index, I noticed that large buffers were being held hostage by the `user_data` array, which doesn't drop user data after it has been returned.

This PR replaces the existing free-list allocator with one that doesn't leak memory. Fortunately, the approach used to avoid memory leaks also ensures that it's safe to not require dummy user data when first constructing a uring, so we can get rid of that.

### DIFF
--- a/lib/uring/heap.ml
+++ b/lib/uring/heap.ml
@@ -1,0 +1,78 @@
+let ( = ) : int -> int -> bool = ( = )
+let ( <> ) : int -> int -> bool = ( <> )
+
+type ptr = int
+let null_ptr = -1
+let is_null x = x = null_ptr
+
+(* Free-list allocator *)
+type 'a t =
+  { data: 'a array
+  (* Pool of potentially-empty data slots. Invariant: an unfreed pointer [p]
+     into this array is valid if one of the following holds:
+
+     1. [is_null free_head];
+     2. [not (is_null free_head)
+         && free_head <> p
+         && is_null free_tail_relation.(p)].
+
+     Either all slots are occupied (1), or there's at least one free slot and
+     [p] is not one of them (2). *)
+  ; mutable free_head: ptr
+  ; free_tail_relation: ptr array
+  (* A linked list of free blocks, with [free_head] being the first element and
+     [free_tail_relation] mapping each free slot [p] to the next one (or
+     [null_ptr] if [p] is either not free or is the final slot).
+
+     Invariant: for all [p : ptr], either [is_null p] or [0 <= p < length]. The
+     user sees non-null pointers only. *)
+  ; max_size: int
+  }
+
+let create : type a. int -> a t =
+ fun n ->
+  if n < 0 || n > Sys.max_array_length then invalid_arg "Heap.create" ;
+  (* Every slot is free, and all but the last have a free successor. *)
+  let free_head = if n = 0 then null_ptr else 0 in
+  let free_tail_relation = Array.init n succ in
+  if n > 0 then free_tail_relation.(n - 1) <- null_ptr;
+  let data =
+    (* See invariants on {!t}, which ensure this data is never read. *)
+    Array.make n (Obj.magic `invalid : a)
+  in
+  { data; free_head; free_tail_relation; max_size = n }
+
+exception No_space
+
+let alloc t a =
+  let ptr = t.free_head in
+  if is_null ptr then raise No_space;
+  let next_head = t.free_tail_relation.(ptr) in
+  Array.unsafe_set t.free_tail_relation ptr null_ptr;
+  Array.unsafe_set t.data ptr a;
+  t.free_head <- next_head;
+  ptr
+
+let free : type a. a t -> ptr -> a =
+ fun t ptr ->
+  let () =
+    assert (not (is_null ptr)) (* [alloc] returns only valid pointers. *);
+    if ptr >= t.max_size then invalid_arg "Heap.free: invalid pointer";
+    let next_free = Array.unsafe_get t.free_tail_relation ptr in
+
+    if is_null t.free_head then
+      (* No free slots, so [ptr] is valid in [t.data]. Case (1) above. *)
+      assert (is_null next_free)
+
+    else if t.free_head <> ptr && is_null next_free then
+      (* One or more free slots, and [ptr] isn't one of them. Case (2) above. *)
+      Array.unsafe_set t.free_tail_relation ptr t.free_head
+
+    else invalid_arg "Heap.free: pointer already freed"
+  in
+  t.free_head <- ptr;
+  let datum = Array.unsafe_get t.data ptr in
+  (* Zero the freed slot. Since this value is now inaccessible (with [ptr] is
+      in the free list) we might as well let it be GCed. *)
+  Array.unsafe_set t.data ptr (Obj.magic `invalid : a);
+  datum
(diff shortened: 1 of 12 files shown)


lib/uring/uring.ml
```diff
@@ -136,16 +129,15 @@ let submit t =
 let errno_is_retry = function -62 | -11 | -4 -> true |_ -> false
 
 let fn_on_ring fn t =
-   let id, res = fn t.uring in
-   match id, res with
+   let (id : Heap.ptr), res = fn t.uring in
+   match (id :> int), res with
    | -1, res when errno_is_retry res ->
```
talex5, line 134: How can `id` (a `Heap.ptr`) be `-1`?

talex5:
Using an array instead of a linked-list to track the free slots seems reasonable. Would be interesting to see some performance numbers.

lib/uring/heap.ml
```diff
@@ -0,0 +1,92 @@
...
+  assert (ptr >= 0) (* [alloc] returns only valid pointers. *);
+  if ptr >= t.length then Fmt.invalid_arg "Heap.free: invalid pointer %d" ptr;
+  let slot_state = Array.unsafe_get t.free_tail_relation ptr in
+  if slot_state <> slot_taken then invalid_arg "Heap.free: pointer already freed";
+
+  (* [t.free_tail_relation.(ptr) = slot_taken], so [t.data.(ptr)] is valid. *)
+  let datum = Array.unsafe_get t.data ptr in
+
+  (* Cons [ptr] to the free-list. *)
+  Array.unsafe_set t.free_tail_relation ptr t.free_head;
```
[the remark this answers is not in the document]
CraigFe: Just a bounds check ([here](https://godbolt.org/z/3Tj3aW598)):

```asm
        ;; Extra check in Array.set
        movq    -8(%rax), %rbx
        cmpq    $2047, %rbx
        jbe     .L108

        ;; Both
        movl    $1, %eax
        addq    $8, %rsp
        ret
```

I've been using it more as a signal to readers than an attempt to extract performance (which I assume is negligible anyway). Happy to drop it.

lib/uring/uring.ml
CraigFe: Yes, this is a hack around the C-stubs wanting to use `-1` as an error code and me wanting to keep IDs `private`. We could just drop the `private` (and change some assertions to errors in `heap.ml`). What do you think?

CraigFe:
Microbenchmark:

```ocaml
let noop_run size =
  let t = Uring.create ~queue_depth:size () in
  Staged.stage (fun () ->
      for i = 1 to size do let r = Uring.noop t i in assert r done;
      let submitted = Uring.submit t in assert (submitted = size);
      for _ = 1 to size do ignore (Uring.wait t : (int * int) option) done)
```

On current `main`:

```
╭──────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name          │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├──────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  noop    10  │             0.0077 mjw/run│           150.0551 mnw/run│           2502.0685 ns/run│
│  noop    30  │             0.0754 mjw/run│           450.1358 mnw/run│           6375.0174 ns/run│
│  noop   100  │             0.8467 mjw/run│          1500.3405 mnw/run│          21088.8327 ns/run│
│  noop   300  │             7.7571 mjw/run│          4500.5976 mnw/run│          63188.8516 ns/run│
│  noop  1000  │            85.7926 mjw/run│         15001.0638 mnw/run│         201523.0778 ns/run│
│  noop  3000  │           794.9051 mjw/run│         45001.8519 mnw/run│         614649.4986 ns/run│
│  noop 10000  │          9265.2980 mjw/run│        150003.3333 mnw/run│        2125486.8748 ns/run│
╰──────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

This PR:

```
╭──────────────┬───────────────────────────┬───────────────────────────┬───────────────────────────╮
│name          │  major-allocated          │  minor-allocated          │  monotonic-clock          │
├──────────────┼───────────────────────────┼───────────────────────────┼───────────────────────────┤
│  noop    10  │             0.0000 mjw/run│           120.0567 mnw/run│           2538.0591 ns/run│
│  noop    30  │             0.0000 mjw/run│           360.1439 mnw/run│           6741.1804 ns/run│
│  noop   100  │             0.0000 mjw/run│          1200.3505 mnw/run│          21781.5844 ns/run│
│  noop   300  │             0.0000 mjw/run│          3600.5792 mnw/run│          59570.2872 ns/run│
│  noop  1000  │             0.0000 mjw/run│         12001.0490 mnw/run│         197404.5607 ns/run│
│  noop  3000  │             0.0000 mjw/run│         36001.8072 mnw/run│         581986.1964 ns/run│
│  noop 10000  │             0.0000 mjw/run│        120003.3333 mnw/run│        1971265.5826 ns/run│
╰──────────────┴───────────────────────────┴───────────────────────────┴───────────────────────────╯
```

So it looks like we had almost exactly 15 minor words per noop cycle before, and now we have 12. It's possible to take it down to 8 by inlining the `with_id` in `Uring.noop`, after which I think the remaining allocations are all options / pairs / closures in `Uring.wait`.

talex5:
Thanks for the benchmark. I tried removing all the `unsafe` and `magic` bits, and it appears to make no difference to the speed:
https://github.com/talex5/ocaml-uring/commit/a240d36f9fe92023b4969b5614b6bab79e55efc4

CraigFe:
I pushed a commit that drops the `unsafe` bits.

talex5:
Nice. Can we get rid of the `magic` too? I know it "should" be impossible for it ever to be returned, but then we thought that about the old code too and you mentioned seeing it return the default value sometimes. It gets pretty hard to debug stuff once the types are wrong (I imagine this could happen if two mutlicore domains tried to the use the same uring, for example).

CraigFe:
I think it's worth avoiding propagating a `~default` parameter throughout user code: there's nothing they can realistically do when it's returned to them, and returning only valid data is already part of the API contract. Needing to pass this parameter is also a limitation to users: I currently have a "request" type that uses a backpointer into the parent structure, and it's not possible to build one of these backpointers w/o first constructing the Uring.

If you're really set against using `Obj` here, I suggest we put options in the slots and take the performance loss. It's similar to the linked list in terms of perf, but keeps the nicer API and fails more gracefully than `Obj`.

talex5:
OK, seems reasonable. This PR needs rebasing though.

CraigFe:
Thanks. Rebased with the Obj removed in favour of options :-)

lib/uring/uring.ml
```diff
@@ -134,27 +133,19 @@ let submit t =
   end else
     0
 
-type 'a completion_option =
+ type 'a completion_option =
```
talex5, line 136:
```suggestion
type 'a completion_option =
```

lib/uring/uring_stubs.c
```diff
@@ -228,6 +229,17 @@ value ocaml_uring_submit(value v_uring)
   CAMLreturn(Val_int(num));
 }
 
+#define Val_cqe_none Val_int(0)
+
+static value Val_cqe_some(value id, value res) {
+  CAMLparam2(id, res);
+  CAMLlocal1(some);
+  some = caml_alloc(3, 0);
```
talex5, line 237:
```suggestion
  some = caml_alloc(2, 0);
```

CraigFe: Oops, thanks.

PR #10, opened 2021-04-23
Add a custom finaliser to Val_uring

CraigFe:
This adds a finaliser to `io_uring` blocks that closes them if they are still open.

I'm not sure how to unit-test this properly, but for now I've added some very weak tests that just ensure no exceptions are raised during finalisation.

### DIFF
--- a/tests/main.ml
+++ b/tests/main.ml
@@ -78,6 +78,32 @@ let test_readv () =
   check_string ~__POS__ ~expected:"est fil" (Bigstringaf.to_string b2);
   ()
 
+let test_finalise () =
+  let () =
+    (* Finalising an unclosed Uring is successful. *)
+    let finalised = ref false in
+    let () =
+      Uring.create ~default:() ~queue_depth:1 ()
+      |> Gc.finalise_last (fun _ -> finalised := true)
+    in
+    Gc.full_major ();
+    assert_ ~__POS__ !finalised;
+    Gc.full_major ()
+  in
+  let () =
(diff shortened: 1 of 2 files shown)


tests/main.ml
```diff
@@ -78,6 +78,32 @@ let test_readv () =
...
+  let () =
+    (* Finalising an unclosed Uring is successful. *)
+    let finalised = ref false in
+    let () =
+      Uring.create ~default:() ~queue_depth:1 ()
+      |> Gc.finalise_last (fun _ -> finalised := true)
+    in
+    Gc.full_major ();
+    assert_ ~__POS__ !finalised;
+    Gc.full_major ()
```
CraigFe, line 91: I have no idea why, but the C finaliser appears to only run on this _second_ forced GC major, even though the first one triggers the OCaml finaliser...

talex5:
Note: there is already a `Gc.finalise Uring.exit uring` in `uring.ml`. Do we still need that?

BTW, I wonder if there is another problem here: e.g.

1. Create a uring.
2. Submit a request to read into the iobuf.
3. GC the ring and iobuf.
4. Linux writes to the iobuf.

Perhaps we should be registering the iobuf as a GC root until the operation is complete?

talex5:
I merged #28 instead, which prevents the ring from being GC'd if `exit` isn't called (since I don't think it's safe to shut down the ring while there are active requests). So I don't think this PR is needed any longer.

PR #13, opened 2021-04-24
Use a more structured completion return type

CraigFe:
The current `(int * 'a) option` return type is a bit error-prone (and allocates an extra pair of words we probably don't need). Perhaps there are better names than "result" and "data" for these components?


talex5 (approved):
Seems like a nice improvement. Can't decide whether naming the new constructors `None` and `Some` will be helpful or confusing, but happy to give it a go!

PR #18, opened 2021-05-25
Don't install test executables

talex5:
They don't seem very useful, and pull in a dependency on Lwt.

@avsm were these installed for a reason?


avsm:
only because i used them for a while as a cp/cat replacement as they were so fast ;-)

PR #19, opened 2021-05-26
Make file offset non-optional

talex5:
In #17 I changed the default offset to `-1`. That worked for files (and pipes!), but it doesn't work for sockets, which require you to pass `0`. As it seems impossible to provide a default that works everywhere, this PR requires the user to specify the offset in all cases.

Also, rename `offset` to `file_offset` in some places for consistency.


avsm:
Looks good (although of course, `file_offset` is totally something else with network sockets, but we can't help what the underlying API does)

PR #20, opened 2021-05-26
Add connect

talex5:
This seems a bit complicated. The problem is that the address is passed as a pointer into application memory, and so must not be GC'd while the request is in progress. So I convert the `Unix.sockaddr` to the C form and store it in a new field I added to @CraigFe's heap so it doesn't get removed until the entry is freed.

Suggestions on better ways of doing it welcome!

(we might have the same problem with the `iovec` arguments)

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -78,7 +78,7 @@ end
 type 'a t = {
   uring: Uring.t;
   mutable fixed_iobuf: Iovec.Buffer.t;
-  user_data : 'a Heap.t;
+  data : 'a Heap.t;
   queue_depth: int;
   mutable dirty: bool; (* has outstanding requests that need to be submitted *)
 }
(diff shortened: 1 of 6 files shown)


CraigFe:
This approach seems good to me :-)

I might have added the box externally, so that `heap.ml` can remain agnostic to the need for this trick:

```ocaml
(* uring.ml *)

(* Create-only handle on an arbitrary OCaml value. *)
module Any : sig
  type t
  val v : 'a -> t
end = struct
  type t = Obj.t
  let v = Obj.repr
end

type 'a t = {
  data : ('a * Any.t) Heap.t;
  ...
}
```

... but this isn't a clear improvement as the pair is not inline as it was before.

lib/uring/uring.ml
```diff
@@ -97,15 +102,17 @@ let realloc t iobuf =
 
 let exit {uring;_} = Uring.exit uring
 
-let with_id : type a. a t -> (Heap.ptr -> bool) -> a -> bool =
- fun t fn datum ->
-  match Heap.alloc t.user_data datum with
+let with_id_full : type a. a t -> (Heap.ptr -> bool) -> a -> extra_data:'b -> bool =
+ fun t fn datum ~extra_data ->
+  match Heap.alloc t.user_data datum ~extra_data with
```
CraigFe, line 107:
```suggestion
  match Heap.alloc t.data datum ~extra_data with
```

Perhaps worth renaming to indicate that this isn't just user data anymore.

talex5: Good idea - done!

PR #25, opened 2021-06-15
Fix test_cancel

talex5:
Occasionally, the read is busy when we try to cancel, giving different results.


talex5:
Also, free the uring after each test, so we don't run out of memory on the test machines.

PR #29, opened 2021-06-22
Replace iovec with cstruct and clean up the C stubs

talex5:
Making iovecs safe was very difficult, because the caller could mutate the array, GC it, or advance the iovecs while the kernel was using the memory. Instead, `readv` and `writev` now take a list of Cstructs (adding a dependency on cstruct). This also allows reading and writing with sub-ranges of bigarrays, and makes it easy to implement an `advance` function that works with multiple buffers.

The C iovec is created inside the io-uring call and not exposed. The C iovec and the list of cstructs is attached to the user-data so it can't be freed until the call completes. The iovec is then freed by the GC like anything else, without an explicit free operation.

Other changes:

- `unregister_bigarray` is now `unregister_buffers`, and the signatures match between the ml and C code.
- Improved handling of OOM errors.
- Document how rings are freed and which functions require the ring to be idle.
- Check path is a valid C string in `ocaml_uring_make_open_how`.
- Don't call `Uring.exit` twice, and make it safe if you do.


lib/uring/uring.ml
```diff
@@ -105,6 +105,31 @@ module Open_how = struct
...
+module Iovec = struct
+  (* The C stubs rely on the layout of Cstruct.t, so we just check here that it hasn't changed. *)
+  module Check : sig
+    [@@@warning "-34"]
+    type t = private {
+      buffer: (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t;
+      off   : int;
+      len   : int;
+    }
+  end = Cstruct
```
avsm, line 117: I like this structural safety check here!

PR #32, opened 2021-06-24
Compile uring using the C flags from OCaml

### LINKED ISSUE #23: Release tracking issue
Things to sort out before release:

- [x] There are a load of `/* TODO sort out cast */` comments in the cstubs (originally added by @avsm) - not sure what needs to be done here.
- [x] Check we never GC things that are still in use (see e.g. #10).
- [x] Sort out iovecs. @avsm doesn't like the current API. Check GC behaviour, alignment TODO.
- [x] `TODO: replace with {!Region.t} instead.` in API. Should be easy enough.

talex5:
Also, remove an unused dune file.

Closes #23.


talex5:
Note sure whether this needs to be `ocamlopt_cflags` for native code. They're both the same on my machine, at least.

avsm:
They're the same.  From utils/config.mlp:

```
let ocamlc_cflags = "%%OCAMLC_CFLAGS%%"
let ocamlc_cppflags = "%%OCAMLC_CPPFLAGS%%"
(* #7678: ocamlopt uses these only to compile .c files, and the behaviour for
          the two drivers should be identical. *)
let ocamlopt_cflags = "%%OCAMLC_CFLAGS%%"
let ocamlopt_cppflags = "%%OCAMLOPT_CPPFLAGS%%"
```

PR #38, opened 2021-10-08
Address fmt strf deprecation error

bikallem:
Context: This project is used as git submodule  in https://github.com/ocaml-multicore/eio. `dune b` seems to fail in a local dev machine due to usage of deprecated functions.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -21,7 +21,7 @@
   (bechamel (and (>= 0.1.0) :with-test))
   (logs (and :with-test (>= 0.5.0)))
   cmdliner
-  (fmt (>= 0.8.4))
+  (fmt (>= 0.8.10))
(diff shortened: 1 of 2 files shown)


talex5:
The opam file says `"fmt" {>= "0.8.4"}`. I don't think `Fmt.str` was in that version, so probably the minimum version will need updating too.

bikallem:
Looks like the `eio` CI runs also fail due to `strf` deprecation. https://ci.ocamllabs.io/github/ocaml-multicore/eio/commit/95cdeafce2dff6f99d6ffdefdaba61fef7ae59c0/variant/debian-11-4.12+domains 

Once this is merged I will update the `eio` PR.

talex5:
Thanks!

PR #49, opened 2022-02-16
Add sendmsg and recvmsg

patricoferris:
This PR adds `sendmsg(2)` and `recvmsg(2)` calls to the uring library. This means having to wrap the `msghdr` struct and I have a few questions.

I have this working with `Eio_linux` when I take the simple UDP test from https://github.com/ocaml-multicore/eio/pull/171 and turn it into an executable. This also seems to work if I place lots of `Gc.{minor,full_major}()` calls in that test. However, as soon as it runs in mdx I get: `Exception: Unix.Unix_error(Unix.EINVAL, "recv_msg", "")`. Note, this is running on WSL. I think something must be being GCed or moved about (although I investigated the msghdr just before this exception is raised and it seemed okay). I'll push the code to Eio after opening this PR . Is there something obviously wrong in the C stubs I wrote perhaps?


lib/uring/uring_stubs.c
```diff
@@ -405,6 +405,46 @@ ocaml_uring_extract_sockaddr(value v) {
...
+// allocate a custom block containing a sendmsg/recvmsg msghdr 
+value
+ocaml_uring_make_msghdr(value v_sockaddr, value v_iov) {
+  CAMLparam2(v_sockaddr, v_iov);
+  CAMLlocal1(v);
+  struct msghdr *msg;
+  struct iovec *iovs = Iovec_val(Field(v_iov, 0));
+  int iovs_len = Int_val(Field(v_iov, 1));
+  // Allocate a pointer on the OCaml heap for the msghdr
+  v = caml_alloc_custom_mem(&msghdr_ops, sizeof(struct msghdr_ops *), sizeof(struct iovec));
```
talex5, line 435:
```suggestion
  v = caml_alloc_custom_mem(&msghdr_ops, sizeof(struct msghdr *), sizeof(struct msghdr));
```

lib/uring/uring_stubs.c
```diff
@@ -405,6 +405,46 @@ ocaml_uring_extract_sockaddr(value v) {
...
+  finalize_msghdr,
+  custom_compare_default,
+  custom_hash_default,
+  custom_serialize_default,
+  custom_deserialize_default,
+  custom_compare_ext_default,
+  custom_fixed_length_default
+};
+
+// allocate a custom block containing a sendmsg/recvmsg msghdr
```
talex5, line 426: This should note that the arguments cannot be freed before the result it, as it contains pointers to them.

lib/uring/uring_stubs.c
```diff
@@ -419,6 +459,32 @@ ocaml_uring_submit_connect(value v_uring, value v_id, value v_fd, value v_sockad
   CAMLreturn(Val_true);
 }
 
+value
```
talex5, line 462: Needs a comment that v_msghdr must not be GC'd while this is running. Same for `ocaml_uring_submit_recv_msg`.

talex5 (commented):
> I think something must be being GCed or moved about (although I investigated the msghdr just before this exception is raised and it seemed okay).

To test this, I suggest doing a `Gc.minor ()` after allocating the memory and just before the call. That will ensure it's on the major heap, and the multicore GC never moves things after that anyway. You're already checking it before the `io_uring_prep_recvmsg` and if you also log a message when it's freed then that rules most things out.

You could also look at the message header just after the error. Has it changed? But I though WSL didn't support io_uring anyway, or has it been updated?

I assume the failing test is the one added by the UDP PR? I can't see anything obviously wrong that would cause that.

patricoferris:
Thanks for the speedy review :))

> To test this, I suggest doing a Gc.minor () after allocating the memory and just before the call. That will ensure it's on the major heap, and the multicore GC never moves things after that anyway. You're already checking it before the io_uring_prep_recvmsg and if you also log a message when it's freed then that rules most things out.

With the executable I tried this and things seemed to be working well.

> But I though WSL didn't support io_uring anyway, or has it been updated?

The very latest release of the [Microsoft Linux kernel supports io_uring](https://github.com/microsoft/WSL2-Linux-Kernel/releases/tag/linux-msft-wsl-5.10.60.1) and all the other tests etc. seem to be working okay. 

> I assume the failing test is the one added by the UDP PR?

Yep that's the one. Interestingly, this (clearly not a fix) patch gets the test in MDX to pass. I think it suggests something is being moved along the way somewhere. I'll keep having a look :))

```diff
diff --git a/lib/uring/uring_stubs.c b/lib/uring/uring_stubs.c
index 0261f0d..a3a7354 100644
--- a/lib/uring/uring_stubs.c
+++ b/lib/uring/uring_stubs.c
@@ -459,30 +459,42 @@ ocaml_uring_submit_connect(value v_uring, value v_id, value v_fd, value v_sockad
   CAMLreturn(Val_true);
 }
 
+struct msghdr msg_send;
+
 // v_msghdr must not be GC'd while the call is in progress
 value
 ocaml_uring_submit_send_msg(value v_uring, value v_id, value v_fd, value v_msgh
dr) {
   CAMLparam2(v_uring, v_msghdr);
   struct io_uring *ring = Ring_val(v_uring);
   struct msghdr *msg = Msghdr_val(Field(v_msghdr, 0));
+  msg_send.msg_name = msg->msg_name;
+  msg_send.msg_namelen = msg->msg_namelen;
+  msg_send.msg_iov = msg->msg_iov;
+  msg_send.msg_iovlen = msg->msg_iovlen;
   struct io_uring_sqe *sqe = io_uring_get_sqe(ring);
   if (!sqe) CAMLreturn(Val_false);
   dprintf("submit_sendmsg\n");
-  io_uring_prep_sendmsg(sqe, Int_val(v_fd), msg, 0);
+  io_uring_prep_sendmsg(sqe, Int_val(v_fd), &msg_send, 0);
   io_uring_sqe_set_data(sqe, (void *)Long_val(v_id));
   CAMLreturn(Val_true);
 }
 
+struct msghdr msg_recv;
+
 // v_msghdr must not be GC'd while the call is in progress
 value
 ocaml_uring_submit_recv_msg(value v_uring, value v_id, value v_fd, value v_msgh
dr) {
   CAMLparam2(v_uring, v_msghdr);
   struct io_uring *ring = Ring_val(v_uring);
   struct msghdr *msg = Msghdr_val(Field(v_msghdr, 0));
+  msg_recv.msg_name = msg->msg_name;
+  msg_recv.msg_namelen = msg->msg_namelen;
+  msg_recv.msg_iov = msg->msg_iov;
+  msg_recv.msg_iovlen = msg->msg_iovlen;
   struct io_uring_sqe *sqe = io_uring_get_sqe(ring);
   if (!sqe) CAMLreturn(Val_false);
   dprintf("submit_recvmsg:msghdr %p: registering iobuf base %p len %lu\n", msg,
 msg->msg_iov[0].iov_base, msg->msg_iov[0].iov_len);
-  io_uring_prep_recvmsg(sqe, Int_val(v_fd), msg, 0);
+  io_uring_prep_recvmsg(sqe, Int_val(v_fd), &msg_recv, 0);
   io_uring_sqe_set_data(sqe, (void *)Long_val(v_id));
   CAMLreturn(Val_true);
 }
```

patricoferris:
I got the CI working on the Eio PR and it gives a different error 😅 

```diff
File "tests/test_network.md", line 1, characters 0-0:
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u tests/test_network.md tests/.mdx/test_network.md.corrected)
diff --git a/tests/test_network.md b/tests/.mdx/test_network.md.corrected
index 95ce3d3..a93244b 100644
--- a/tests/test_network.md
+++ b/tests/.mdx/test_network.md.corrected
@@ -174,8 +174,7 @@ val pp_addr :
       Eio.Net.send e e2 (Cstruct.of_string "UDP Message"));;
 +Waiting to receive data on udp:127.0.0.1:8082
 +Sending data from udp:127.0.0.1:8081 to udp:127.0.0.1:8082
-+Received message from 127.0.0.1:8081: UDP Message
-- : unit = ()
+Exception: Unix.Unix_error(Unix.ENOBUFS, "send_to", "")
 ```
 

```

patricoferris:
Ah! I think I need to memset the msghdr to `0` or at the very least set `msg_control, msg_controllen, msg_flags` to `0`. Trying that now 🤞

PR #51, opened 2022-02-17
Add an io_uring probe function

patricoferris:
This PR adds a `Uring.probe ()` function that checks the kernel's support for the various io_uring operations that we currently use (I think I got all of them). Hopefully this can help with https://github.com/ocaml-multicore/eio/issues/187 and could be used in `Eio_main` for Linux to check if we can using the io_uring backend before defaulting to libuv.

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -137,6 +137,7 @@ module Uring = struct
 
   external create : int -> int option -> t = "ocaml_uring_setup"
   external exit : t -> unit = "ocaml_uring_exit"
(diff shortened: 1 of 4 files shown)


talex5 (changes requested):
I don't think commenting out the strings like this will work; `io_uring_opcode_supported` only sees the index, not the string.

Also, while this function is useful for Eio at the moment (which can switch to luv if anything is missing), in general users will want to probe individual features themselves. As new calls get added, Eio might want to use them only if available.

So probably we should add the `IO_URING_OP_*` values to `lib/uring/include/discover.ml` and expose them in the interface.

PR #53, opened 2022-02-25
Don't allocate a fixed buffer by default

talex5:
Instead, let the user call `set_fixed_buffer` manually if they want one. This avoids allocating locked memory by accident, and makes it easy to handle the case of hitting a resource limit.

/cc @avsm


avsm (commented):
Looks good to me.

PR #59, opened 2022-07-27
Allow clock to be configured in Uring.wait

bikallem:
This PR allows specifying clocks for timeout (as timeout is relative to a clock) when waiting for `cqe` events; specifically `Uring.wait` now accepts clock argument along with the `timeout` value in nanoseconds.

Summary:
- Changes timeout from seconds to nanoseconds (float -> int64)
- Add `Clock_mono`, `Clock_sys` ~~and `Clock_default` to `Uring.wait`~~

Required by https://github.com/ocaml-multicore/eio/pull/249

### DIFF
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -67,11 +67,14 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
-val timeout: ?absolute:bool -> 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
-(** [timeout t clock ns d] submits a timeout request to uring [t].
+(** {2 Timeout} *)
 
-    [clock] [`Boottime] and [`Realtime] represents OS clocks CLOCK_BOOTTIME and CLOCK_REALTIME 
-    respectively.
+type clock = Boottime | Realtime
+(** Represents Linux clocks. [Boottime] and [Realtime] represents OS clocks CLOCK_BOOTTIME
+    and CLOCK_REALTIME respectively. *)
+
+val timeout: ?absolute:bool -> 'a t -> clock -> int64 -> 'a -> 'a job option
+(** [timeout t clock ns d] submits a timeout request to uring [t].
 
     [absolute] denotes how [clock] and [ns] relate to one another. Default value is [false]
 
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -67,13 +67,13 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
-val timeout: ?rel:[`Relative | `Absolute] -> 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
+val timeout: ?absolute:bool -> 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
 (** [timeout t clock ns d] submits a timeout request to uring [t].
 
     [clock] [`Boottime] and [`Realtime] represents OS clocks CLOCK_BOOTTIME and CLOCK_REALTIME 
     respectively.
 
-    [rel] denotes how [clock] and [ns] relates to one another. Default value is [`Relative]
+    [absolute] denotes how [clock] and [ns] relate to one another. Default value is [false]
 
     [ns] is the timeout time in nanoseconds *)
 
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -67,6 +67,14 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
+val timeout: 'a t -> ?clock:[`Clock_mono | `Clock_sys | `Clock_default] -> int64 -> 'a -> 'a job option
+(** [timeout t d clock ns] submits a timeout request to uring [t].
+
+    [clock] [`Clock_default] represents the default clock used by uring, [`Clock_mono] represents 
+    CLOCK_BOOTTIME and [`Clock_sys] represents CLOCK_REALTIME.
+
+    [ns] is the the timeout time in nanoseconds *)
+
 module type FLAGS = sig
   type t = private int
   (** A set of flags. *)
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -67,13 +67,15 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
-val timeout: 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
+val timeout: ?rel:[`Relative | `Absolute] -> 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
 (** [timeout t clock ns d] submits a timeout request to uring [t].
 
-    [clock] [`Boottime] and [`Realtime] represents CLOCK_BOOTTIME and
-    CLOCK_REALTIME respectively.
+    [clock] [`Boottime] and [`Realtime] represents OS clocks CLOCK_BOOTTIME and CLOCK_REALTIME 
+    respectively.
 
-    [ns] is the the timeout time in nanoseconds *)
+    [rel] denotes how [clock] and [ns] relates to one another. Default value is [`Relative]
+
+    [ns] is the timeout time in nanoseconds *)
 
 module type FLAGS = sig
   type t = private int
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -67,11 +67,11 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
-val timeout: 'a t -> ?clock:[`Clock_mono | `Clock_sys | `Clock_default] -> int64 -> 'a -> 'a job option
-(** [timeout t d clock ns] submits a timeout request to uring [t].
+val timeout: 'a t -> [`Boottime | `Realtime] -> int64 -> 'a -> 'a job option
+(** [timeout t clock ns d] submits a timeout request to uring [t].
 
-    [clock] [`Clock_default] represents the default clock used by uring, [`Clock_mono] represents 
-    CLOCK_BOOTTIME and [`Clock_sys] represents CLOCK_REALTIME.
+    [clock] [`Boottime] and [`Realtime] represents CLOCK_BOOTTIME and
+    CLOCK_REALTIME respectively.
 
     [ns] is the the timeout time in nanoseconds *)
 
--- a/lib/uring/uring_stubs.c
+++ b/lib/uring/uring_stubs.c
@@ -33,7 +33,6 @@
 #include <string.h>
 #include <poll.h>
 #include <sys/uio.h>
-#include <stdint.h>
 
 #undef URING_DEBUG
 #ifdef URING_DEBUG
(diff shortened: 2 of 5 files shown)


talex5 (commented):
I think the wait API should match the C library. You should be able to add `io_uring_prep_timeout` as a separate operation, like the other jobs.

bikallem:
> I think the wait API should match the C library. You should be able to add `io_uring_prep_timeout` as a separate operation, like the other jobs.

Does that mean we need another version of `wait_cqe_timeout` - for eg. `wait_cqe_timeout2`. Something like this:
```
let wait ?clock ?timeout t =
  match timeout with
  | None -> fn_on_ring Uring.wait_cqe t
  | Some timeout -> begin
    assert(timeout >= 0L);
    match clock with 
    | Some c -> fn_on_ring (Uring.wait_cqe_timeout2 c timeout) t
    | None ->  fn_on_ring (Uring.wait_cqe_timeout timeout) t
```

talex5:
There shouldn't be any need to change anything about `wait` to support different clocks. Just add a new function for timeouts. Something like:

```ocaml
val timeout : 'a t -> [`Clock_mono |`Clock_sys | `Clock_default] -> int64 -> 'a -> 'a job option
```

Then users can submit (and cancel) timeout operations just like any other job.

bikallem:
@talex5 created a new function `timeout` as suggested.

lib/uring/uring.mli
```diff
@@ -67,6 +67,14 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
+val timeout: 'a t -> ?clock:[`Clock_mono | `Clock_sys | `Clock_default] -> int64 -> 'a -> 'a job option
```
talex5, line 70: This is a Linux-only library (not a cross-platform abstraction), so the names should match the ones in the C API:
```suggestion
val timeout: 'a t -> clock:[`Boottime|`Realtime] -> int64 -> 'a -> 'a job option
```

lib/uring/uring_stubs.c
```diff
@@ -184,6 +185,40 @@ ocaml_uring_submit_nop(value v_uring, value v_id) {
...
+    if (res < 0)
+      CAMLreturn(Val_false);
+
+    sqe = io_uring_get_sqe(ring);
+    if(!sqe)
+      CAMLreturn(Val_false);
+  }
+  io_uring_prep_timeout(sqe, &t, 0, clock);
+  io_uring_sqe_set_data(sqe, (void *)Long_val(v_id));
+  sqe->user_data = LIBURING_UDATA_TIMEOUT;
```
talex5, line 218: What is this for?

lib/uring/uring_stubs.c
```diff
@@ -184,6 +185,40 @@ ocaml_uring_submit_nop(value v_uring, value v_id) {
...
+  else if (v_clock == caml_hash_variant("Clock_sys"))
+    clock = IORING_TIMEOUT_REALTIME;
+  else
+    clock = 0;
+
+  t.tv_sec = 0;
+  t.tv_nsec = Int64_val(v_timeout);
+
+  sqe = io_uring_get_sqe(ring);
+  if(!sqe) {
```
talex5, line 207: Auto-submitting if the ring is full might be useful in future, but it needs more thinking about. For now, just return false like the other operations do and we can find a general solution later.

lib/uring/uring_stubs.c
```diff
@@ -184,6 +185,40 @@ ocaml_uring_submit_nop(value v_uring, value v_id) {
   CAMLreturn(Val_true);
 }
 
+value ocaml_uring_submit_timeout(value v_uring, value v_id, value v_clock, value v_timeout)
+{
+  CAMLparam4(v_uring, v_id, v_clock, v_timeout);
```
talex5, line 190: This should be `noalloc` now (PR needs rebasing; see #61).

lib/uring/uring_stubs.c
```diff
@@ -184,6 +185,40 @@ ocaml_uring_submit_nop(value v_uring, value v_id) {
...
+  struct __kernel_timespec t;
```
talex5, line 191: You can't pass a stack-allocated timeout. It needs to remain alive until submit time. @haesbaert is working on a PR to make this easier, I think.

talex5 (commented):
Needs a test-case added to `tests/main.md`.

lib/uring/uring_stubs.c
haesbaert: Ack, I have a silly allocator that can be used for this., it need a bit more work as I want to maybe rely on Uring.submit being called more promiscuously, which would simplify things a lot.

lib/uring/uring_stubs.c
bikallem: Do you mean `sqe->user_data = LIBURING_UDATA_TIMEOUT;` ?

It follows https://github.com/axboe/liburing/blob/02adcf497c9f4ca20175b291f8269faff4da958c/src/queue.c#L278

lib/uring/uring_stubs.c
talex5: Yes: allocate some memory, store a reference to it somewhere so it can't get GC'd, pass a pointer to it to Linux, then drop the reference at some point after submitting (or after completing, which is what the iovec stuff currently does).

bikallem:
@talex5 `Uring.submit` on `Uring.timeout` job returns `-62`. Is this optimal/ergonomic? Should it perhaps return `[ETIME | `Int of int]`? Or perhaps a helper function `val is_timeout : int -> bool` to decode the return value from `Uring.submit`?

lib/uring/uring_stubs.c
```diff
@@ -146,6 +148,34 @@ ocaml_uring_sq_ready(value v_uring) {
...
+  int clock;
+
+  if (v_clock == caml_hash_variant("Boottime"))
+    clock = IORING_TIMEOUT_BOOTTIME;
+  else
+    clock = IORING_TIMEOUT_REALTIME;
+
+  sqe = io_uring_get_sqe(ring);
+  if (!sqe) return Val_false;
+  io_uring_prep_timeout(sqe, t, 0, clock);
```
talex5, line 174: > Uring.timeout job returns -62. Is this optimal/ergonomic?

```suggestion
  io_uring_prep_timeout(sqe, t, 0, clock | IORING_TIMEOUT_ETIME_SUCCESS);
```
It would be good to support `IORING_TIMEOUT_ABS` too (as a flag).

talex5 (commented):
Looks good to me, once the flag(s) are added.

lib/uring/uring_stubs.c
bikallem: Done.

lib/uring/uring.mli
```diff
@@ -67,6 +67,15 @@ val noop : 'a t -> 'a -> 'a job option
 (** [noop t d] submits a no-op operation to uring [t]. The user data [d] will be
     returned by {!wait} or {!peek} upon completion. *)
 
+val timeout: 'a t -> [`Boottime | `Realtime | `Timeout_abs] -> int64 -> 'a -> 'a job option
```
talex5, line 70:
```suggestion
val timeout: 'a t -> [`Boottime | `Realtime] -> [`Relative|`Absolute] -> int64 -> 'a -> 'a job option
```
`IORING_TIMEOUT_ABS` is a flag, not a clock.

lib/uring/uring.mli
```diff
@@ -67,6 +67,15 @@ val noop : 'a t -> 'a -> 'a job option
...
+(** [timeout t clock ns d] submits a timeout request to uring [t].
+
+    [clock] [`Boottime] and [`Realtime] represents OS clocks CLOCK_BOOTTIME and CLOCK_REALTIME 
+    respectively denoting [ns] as a value relative to the specified clock. [`Timeout_abs] specifies [ns]
+    as an absolute value rather than a relative one.
+
+    [ns] is the the timeout time in nanoseconds *)
```
talex5, line 77:
```suggestion
    [ns] is the timeout time in nanoseconds *)
```

tests/main.md
```diff
@@ -748,3 +748,25 @@ val b : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 1}
 # Uring.exit t;;
 - : unit = ()
 ```
+
+## Timeout
+
+Timeout should return (-ETIME). This is defined in https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/errno.h#L45
```
talex5, line 754: This comment is now wrong.

lib/uring/uring_stubs.c
bikallem: using `IORING_TIMEOUT_ETIME_SUCCESS` seems to give EINVAL (-22). https://ci.ocamllabs.io/github/ocaml-multicore/ocaml-uring/commit/41f87340200483507ce5050f1a6ba86e67ab47c3/variant/debian-11-4.12_opam-2.1

Perhaps the ocaml-ci is not using the supported version of uring. In my local test, even after using `IORING_TIMEOUT_TIME_SUCCESS` it still returns -65.

talex5: It looks like `IORING_TIMEOUT_ETIME_SUCCESS` doesn't change the result value, it just doesn't mark the job as failed for chaining. And also requires Linux 5.16. So we should revert this change and let the caller handle it.

lib/uring/uring.mli
bikallem: Not sure what is happening but when I use `IORING_TIMEOUT_ABS` both as a flag or the only clock source, the timeout is never armed, i.e. `flags |= IORING_TIMEOUT_ABS`. My github and google-fu for IORING_TIMEOUT_ABS doesn't bring that much clarity. I think we should drop this flag for now and add it in the future PR. I don't immediately have need of this flag for now.

lib/uring/uring_stubs.c
```diff
@@ -146,6 +148,37 @@ ocaml_uring_sq_ready(value v_uring) {
...
+
+value /* noalloc */
+ocaml_uring_submit_timeout(value v_uring, value v_id, value v_sketch_ptr, value v_clock, value v_rel)
+{
+  struct __kernel_timespec *t = Sketch_ptr_val(v_sketch_ptr);
+  struct io_uring* ring = Ring_val(v_uring);
+  struct io_uring_sqe* sqe;
+  int flags;
+
+  if (v_clock == caml_hash_variant("Boottime"))
```
talex5, line 167: It looks like `caml_hash_variant` does some calculation. I'm not sure the C compiler will evaluate it at compile time. Maybe this should be calculated once as a constant?

tests/main.md
```diff
@@ -748,3 +748,35 @@ val b : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 1}
...
+
+```ocaml
+# let t = Uring.create ~queue_depth:1 ();;
+val t : '_weak13 Uring.t = <abstr>
+
+# let ns = Int64.(mul 1L 1_000_000_000L) in
```
talex5, line 760: It would be better to use smaller timeouts here (e.g. 10ms instead of 1s). The tests took about 400ms to run before. Adding these tests makes it 6 times slower.

bikallem: Ack.

lib/uring/uring_stubs.c
bikallem: Is this not working as expected? My printf debuggging on this conversion seems to suggest that the values are correct, i.e. `Boottime -> IORING_TIMEOUT_BOOTTIME`, `Absolute -> IORING_TIMEOUT_ABS`. 

Another approach I tried earlier was to define these values statically using dune-configurator. However that didn't work as there is an existing dune issue where the dune-configurator can't include the vendored "liburing.h" header file. (https://ocamllabs.slack.com/archives/C16082YMP/p1658838921704329).

tests/main.md
bikallem: Done.

lib/uring/uring_stubs.c
talex5: It works. i just thought it might be quicker to life this out of the function, and have something like:
```suggestion
  if (v_clock == variant_boottime))
```
But it might be fast enough that it doesn't matter, or it might be that the compiler evaluates it at compile time anyway.

bikallem: There is now a regular variant `type clock = Boottime | Realtime` defined for the clocks. This allows us to use `#define Val_boottime Val_int(0)` which should help with performance.

talex5 (approved):

PR #61, opened 2022-08-02
Be more consistent with CAMLparam

haesbaert:
`v_fileoff` is a boxed value in 32bit systems, so it should be under CAMLparam.
While here fix the same thing in another couple of places and consistently use
`v_fileoff` when it refers to the Int63.t file_offset.

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -159,7 +159,7 @@ module Uring = struct
 
   external unregister_buffers : t -> unit = "ocaml_uring_unregister_buffers"
   external register_bigarray : t ->  Cstruct.buffer -> unit = "ocaml_uring_register_ba"
-  external submit : t -> int = "ocaml_uring_submit"
+  external submit : t -> int = "ocaml_uring_submit" [@@noalloc]
 
   type id = Heap.ptr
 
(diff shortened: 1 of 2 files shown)


talex5:
The old code is certainly inconsistent. But now I wonder if we need `CAMLparam` at all for most of these. I guess anything marked as `[@@noalloc]`  in `uring.ml` shouldn't need to be registering GC roots at all (https://blog.janestreet.com/faster-ocaml-to-c-calls/, https://v2.ocaml.org/manual/intfc.html#ss:c-direct-call).

haesbaert:
That makes more sense tbh, none of them block so they can all be @@noalloc, in that case should i just zap every CAMLparam ?

talex5:
> in that case should we just zap every CAMLparam?

For all the noalloc ones, yes, I think so.

talex5 (commented):
Looks right, but there are a couple of other places that look like they could be noalloc too. If not, could do with a comment saying why.

haesbaert:
zefixed.
I also had forgotten to mush the .ml change for Uring.submit, should be fine now

talex5 (approved):

PR #62, opened 2022-08-03
Readwrite

haesbaert:
Implement Uring.read and Uring.write

    These are simple wrappers for read(2) and write(2) that use a Cstruct.t.
    This avoids the extra allocations done by readv(2) and writev(2) as we currently
    need to do one malloc per readv(2)/writev(2).

    It's also more natural for most applications that only have a single buffer to
    read/write.

+++

    YACI - Yet Another Cat Implementation

    Wrote this to make sure I realize what was going on in Uring and to be able to
    properly test a new allocator for writev/readv, as a bonus I've used to make
    sure read and write work.

    The current urcat is a bit limited and this works a bit more like a normal cat.

### DIFF
--- a/tests/main.md
+++ b/tests/main.md
@@ -343,18 +343,20 @@ val b2_len : int = 7
 # let b1 = Cstruct.create b1_len and b2 = Cstruct.create b2_len;;
 val b1 : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 3}
 val b2 : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 7}
+
 # Uring.read t fd b1 `Read ~file_offset:Int63.minus_one;;
 - : [ `Read ] Uring.job option = Some <abstr>
-# Uring.read t fd b2 `Read ~file_offset:Int63.minus_one;;
-- : [ `Read ] Uring.job option = Some <abstr>
-
 # Uring.submit t;;
-- : int = 2
-
+- : int = 1
 # let `Read, read = consume t;;
 val read : int = 3
 # Cstruct.to_string b1;;
 - : string = "A t"
+
+# Uring.read t fd b2 `Read ~file_offset:Int63.minus_one;;
+- : [ `Read ] Uring.job option = Some <abstr>
+# Uring.submit t;;
+- : int = 1
 # let `Read, read = consume t;;
 val read : int = 7
 # Cstruct.to_string b2;;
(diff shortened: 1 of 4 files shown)


tests/main.md
```diff
@@ -329,11 +329,85 @@ val len : int = 5
...
+# let b1_len = 3 and b2_len = 7;;
+val b1_len : int = 3
+val b2_len : int = 7
+# let b1 = Cstruct.create b1_len and b2 = Cstruct.create b2_len;;
+val b1 : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 3}
+val b2 : Cstruct.t = {Cstruct.buffer = <abstr>; off = 0; len = 7}
+# Uring.read t fd b1 `Read ~file_offset:Int63.minus_one;;
+- : _[> `Read ] Uring.job option = Some <abstr>
+# Uring.read t fd b2 `Read ~file_offset:Int63.minus_one;;
+- : _[> `Read ] Uring.job option = Some <abstr>
```
talex5, line 349: This isn't right, I think. uring doesn't promise to execute operations in order by default, though there is a link flag (`IOSQE_IO_LINK`) that can be used for this. Specifying the file offsets would fix it.

tests/main.md
```diff
@@ -329,11 +329,85 @@ val len : int = 5
...
+
+# Uring.write t w wb `Write ~file_offset:Int63.minus_one;;
+- : _[> `Write ] Uring.job option = Some <abstr>
+# Uring.read t r rb `Read ~file_offset:Int63.minus_one;;
+- : _[> `Read | `Write ] Uring.job option = Some <abstr>
+
+# Uring.submit t;;
+- : int = 2
+
+# let v, read = consume t;;
```
talex5, line 388: Although likely, I'm not sure the completions are certain to come in order. See e.g.

https://github.com/ocaml-multicore/ocaml-uring/blob/c872cd6a381344eaa30acca65a77b516c1eb68a2/tests/main.md?plain=1#L450-L455

tests/ring_cat.ml
```diff
@@ -0,0 +1,100 @@
...
+  let _ = Uring.submit t in
+
+  let next_read_offset file_offset = Int63.(add file_offset (of_int io_size)) in
+
+  let rec loop read_offset goteof pending : unit =
+    if pending = 0 then
+      ()
+    else
+      match Uring.wait t with
+      | None -> failwith "Wait got None"
```
talex5, line 52: Could happen I think if e.g. the terminal is resized.

tests/ring_cat.ml
```diff
@@ -0,0 +1,100 @@
...
+      | Some c ->
+        (* epr "c.result = %d\n%!" c.result; *)
+        if c.result < 0 then
+          failwith "io error :-("
```
talex5, line 56:
```suggestion
          raise Unix.Unix_error(Uring.error_of_errno c.result, "read/write", "")
```

tests/ring_cat.ml
```diff
@@ -0,0 +1,100 @@
...
+        else if c.result = 0 then
+          loop read_offset true (pred pending)
+        else
+          match c.data with
+          | `Read (buf, file_offset) ->
+            (* epr "read %d bytes\n" c.result; *)
+            if goteof then
```
talex5, line 63: Shouldn't we write out the block we read in all cases?

talex5 (commented):
Thanks. Not quite convinced that cat will handle various errors correctly.

I think a short read should instead restart the read on the remainder of the block (Linux doesn't normally do short reads on files, but I think I've seen it in some cases where part of the file is in the cache and part isn't).

haesbaert:
I have dropped the ringcat until I figure out the proper semantics of non seekable FDs.
I have force pushed this and addressed your points on different commits.

talex5 (approved):
Thanks! Looks ready to merge, once CI passes.

PR #64, opened 2022-08-17
Add a readv benchmark and fix sketch buffer GC

talex5:
The benchmark performs 16 concurrent reads of /dev/zero in a loop. Each time a read completes, it issues a new one.

This benchmark unconvered some problems in the current sketch buffer handling:

- A typo in the C stubs meant that the sketch buffer was never released.

- Even with that fixed it might still never be freed, as the application is not required to call submit explicitly. Instead, we now try to release the buffer when waiting or peeking.

- I set the initial sketch buffer size to 4096. Before, it was zero, which lead to quite a lot of allocation as it grows rather slowly. I'm not sure if this really matters, however.

On my machine, I get:

    Read 95.37 MB in 0.25 seconds (388.92 MB/second) # buffer_size=100, polling=false
    Read 95.37 MB in 0.47 seconds (203.45 MB/second) # buffer_size=100, polling=true
    Read 95.37 MB in 0.24 seconds (389.73 MB/second) # buffer_size=100, polling=false
    Read 95.37 MB in 0.48 seconds (200.75 MB/second) # buffer_size=100, polling=true

With a larger buffer:

    Read 124998.12 MB in 5.11 seconds (24452.64 MB/second) # buffer_size=131072, polling=false
    Read 124999.76 MB in 5.14 seconds (24318.08 MB/second) # buffer_size=131072, polling=true
    Read 124993.41 MB in 4.95 seconds (25236.34 MB/second) # buffer_size=131072, polling=false
    Read 124999.46 MB in 5.17 seconds (24157.28 MB/second) # buffer_size=131072, polling=true

I'm not sure why polling mode is slower. Possibly because reading from /dev/zero doesn't benefit from happing asynchronously.

### DIFF
--- a/bench/readv.ml
+++ b/bench/readv.ml
@@ -5,8 +5,6 @@ let buffer_size = 100   (* Use a small buffer to stress the system more *)
 let n_concurrent = 16   (* How many requests to have active at once *)
 let n_iters = 1_000_000 (* How many times to accept and resubmit *)
 
-let got = ref 0
-
 let rec wait t handle =
   match Uring.peek t with
   | Some { result; data = buf } -> handle result buf
@@ -16,7 +14,7 @@ let rec wait t handle =
     | Some { result; data = buf } -> handle result buf
 
 let run_bechmark ~polling_timeout fd =
-  got := 0;
+  let got = ref 0 in
   (* For polling mode, [queue_depth] needs to be slightly larger than [n_concurrent] or submission
      occasionally fails for some reason. *)
   let t = Uring.create ?polling_timeout ~queue_depth:(n_concurrent * 2) () in
@@ -40,12 +38,15 @@ let run_bechmark ~polling_timeout fd =
         )
       )
   done;
+  (* Get a snapshot of the stats before letting things finish. *)
+  let stats = Uring.get_debug_stats t in
   let t1 = Unix.gettimeofday () in
   let time = t1 -. t0 in
   let got = float !got /. (1024. *. 1024.)  in
   let polling = polling_timeout <> None in
-  Printf.printf "Read %.2f MB in %.2f seconds (%.2f MB/second) # buffer_size=%d, polling=%b\n%!"
-    got time (got /. time) buffer_size polling;
+  Fmt.pr "@[<v2>Read %.2f MB in %.2f seconds (%.2f MB/second) # buffer_size=%d, polling=%b@,%a@]@."
+    got time (got /. time) buffer_size polling
+    Uring.Stats.pp stats;
   (* Finally, drain the remaining reads and shut down the ring. *)
   for _ = 1 to n_concurrent do
     wait t (fun _result _buf -> ())
--- a/bench/readv.ml
+++ b/bench/readv.ml
@@ -1,0 +1,61 @@
+(* This benchmark uses [Uring.readv] to read from /dev/zero in a loop.
+   This is intended to stress the sketch buffer. *)
+
+let buffer_size = 100   (* Use a small buffer to stress the system more *)
+let n_concurrent = 16   (* How many requests to have active at once *)
+let n_iters = 1_000_000 (* How many times to accept and resubmit *)
+
+let got = ref 0
+
+let rec wait t handle =
+  match Uring.peek t with
+  | Some { result; data = buf } -> handle result buf
+  | None ->
+    match Uring.wait t with
+    | None -> wait t handle
+    | Some { result; data = buf } -> handle result buf
+
+let run_bechmark ~polling_timeout fd =
+  got := 0;
+  (* For polling mode, [queue_depth] needs to be slightly larger than [n_concurrent] or submission
+     occasionally fails for some reason. *)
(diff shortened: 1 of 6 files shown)


lib/uring/uring_stubs.c
```diff
@@ -143,7 +143,7 @@ ocaml_uring_submit_nop(value v_uring, value v_id) {
 value /* noalloc */
 ocaml_uring_sq_ready(value v_uring) {
   struct io_uring *ring = Ring_val(v_uring);
-  return (Int_val(io_uring_sq_ready(ring)));
```
haesbaert, line 146: :anguished: my bad :disappointed:

lib/uring/uring.ml
```diff
@@ -413,17 +414,24 @@ let cancel t job user_data =
...
-    if t.dirty then begin
-      t.dirty <- false;
-      Uring.submit t.uring
-    end else
-      0
-  in
-  if Uring.sq_ready t.uring = 0 then
-    Sketch.release t.sketch;
-  v
+  if t.dirty then begin
```
haesbaert, line 425: The rationale for checking `sq_ready` even if `dirty` was not set, is that technically you can still allocate things and never have `dirty` set, also more as of a safe guard in the future, checking `sq_ready` is cheap enough.

lib/uring/uring.ml
```diff
@@ -148,6 +148,7 @@ module Sketch = struct
     if alloc_len > avail t then begin
       let new_buffer = create_buffer ((length t) + alloc_len) in
       t.old_buffers <- t.buffer :: t.old_buffers;
+      (* Printf.printf "len t.old_buffers = %d\n%!" (List.length t.old_buffers); *)
```
haesbaert, line 151: debug leftover

bench/readv.ml
```diff
@@ -0,0 +1,60 @@
+(* This benchmark uses [Uring.readv] to read from /dev/zero in a loop.
+   This is intended to stress the sketch buffer. *)
+
+let buffer_size = 100   (* Use a small buffer to stress the system more *)
```
haesbaert, line 4: As a note, the performance drop relevant to buffer size is consistent to what I observed in my old `catv`, 128k was the sweet spot for me, it's also what GNU's cat was using.

bench/readv.ml
```diff
@@ -0,0 +1,60 @@
...
+  done;
+  assert (Uring.submit t = n_concurrent);
+  (* Then we wait for reads to complete. Each time a read returns, we immediately submit another. *)
+  let t0 = Unix.gettimeofday () in
+  for _ = 1 to n_iters do
+    wait t (fun result bufs ->
+        if result < 0 then (
+          raise (Unix.Unix_error (Uring.error_of_errno result, "readv", ""))
+        ) else (
+          got := !got + result;
```
haesbaert, line 37: I'd assert that `result = buffer_size`, if/when someone changes buffersize to one brazilion, it might return less.

bench/readv.ml
```diff
@@ -0,0 +1,60 @@
...
+    wait t (fun _result _buf -> ())
+  done;
+  Uring.exit t
+
+let () =
+  let fd = Unix.openfile "/dev/zero" Unix.[O_RDONLY] 0 in
+  run_bechmark fd ~polling_timeout:None;
+  run_bechmark fd ~polling_timeout:(Some 1000);
+  run_bechmark fd ~polling_timeout:None;
+  run_bechmark fd ~polling_timeout:(Some 1000);
```
haesbaert, line 60: `Unix.close` on fd (sorry for nitpicking)

lib/uring/uring.ml
```diff
@@ -288,7 +289,7 @@ let create ?polling_timeout ~queue_depth () =
   let data = Heap.create queue_depth in
   let id = object end in
   let fixed_iobuf = Cstruct.empty.buffer in
-  let sketch = Sketch.create 0 in
+  let sketch = Sketch.create 4096 in
```
haesbaert, line 292: I kinda liked the starting with no resources and engorging, but that's bikeshedding, I'm fine with either.

haesbaert (changes requested):

lib/uring/uring.ml
talex5: OK, I've put it back how it was before.

lib/uring/uring.ml
talex5: OK, I've removed this and added a new `Uring.Stats` module, which allows you to get some stats about how the ring and sketch buffer are behaving.

bench/readv.ml
talex5: That wouldn't be an error, though, and we report the actual size we read so the results will still be accurate.

bench/readv.ml
talex5: OK, done!

lib/uring/uring.ml
talex5: OK. I've changed it to start from zero again, but it now grows faster.

haesbaert (approved):

haesbaert:
Awesome ! thanks for the fixes \o/

PR #65, opened 2022-08-22
Add unlink

talex5:
This can also be used as rmdir.

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -175,6 +175,15 @@ module Sketch = struct
       set ptr csl;
       ptr
   end
+
+  module String = struct
+    external set : ptr -> string -> unit = "ocaml_uring_set_string" [@@noalloc]
+
+    let alloc t s =
+      let ptr = alloc t (String.length s + 1) in
+      set ptr s;
+      ptr
+  end
 end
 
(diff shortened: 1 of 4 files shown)


haesbaert (approved):
I was gonna suggest to use `memcpy` thinking `String_val` was not guaranteed to be terminated, but it is.

talex5:
> I was gonna suggest to use memcpy thinking String_val was not guaranteed to be terminated, but it is.

That's not a bad idea. At the moment, we stop copying at the first `\0`, which is OK for paths but maybe not for strings in general.

However, the sketch buffer currently sets the length to the rounded length, not the length requested, so that won't work unless we change it to use the requested length.

haesbaert:
> > I was gonna suggest to use memcpy thinking String_val was not guaranteed to be terminated, but it is.
> 
> That's not a bad idea. At the moment, we stop copying at the first `\0`, which is OK for paths but maybe not for strings in general.
> 

I considered that, but we're calling it a string, and you're using it a char *, and it should be ok for every C string, as long as it's a real string.

> However, the sketch buffer currently sets the length to the rounded length, not the length requested, so that won't work unless we change it to use the requested length.

True, but if what we needed was not a string, the memcpy would be enough to not get a truncated thing, which would be the bug, if the area allocated is larger than the memcpy'ed, it's all fine.

PR #70, opened 2022-08-24
Add support for uring probes

talex5:
Closes #51. /cc @patricoferris

### DIFF
--- a/lib/uring/dune
+++ b/lib/uring/dune
@@ -12,6 +12,11 @@
 
 (rule
  (targets config.ml)
+ (deps
+   include/liburing.h
+   include/liburing/io_uring.h
+   include/liburing/barrier.h
+   include/liburing/compat.h)
  (action (run ./include/discover.exe)))
 
 (rule
--- a/lib/uring/include/discover.ml
+++ b/lib/uring/include/discover.ml
@@ -2,49 +2,122 @@ module C = Configurator.V1
 
 let () =
   C.main ~name:"discover" (fun c ->
-    C.C_define.import c ~c_flags:["-D_GNU_SOURCE"]
-      ~includes:["fcntl.h"; "poll.h"; "sys/uio.h"; "linux/time_types.h"]
-      C.C_define.Type.[
-          "POLLIN", Int;
-          "POLLOUT", Int;
-          "POLLERR", Int;
-          "POLLHUP", Int;
+      let defs =
+        C.C_define.import c ~c_flags:["-D_GNU_SOURCE"; "-I"; Filename.concat (Sys.getcwd ()) "include"]
+          ~includes:["fcntl.h"; "poll.h"; "sys/uio.h"; "liburing.h"]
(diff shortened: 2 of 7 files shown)


lib/uring/dune
```diff
@@ -12,7 +12,14 @@
...
 (rule
  (targets config.ml)
- (action (run ./include/discover.exe)))
+ (deps
+   include/liburing.h
+   include/liburing/io_uring.h
+   include/liburing.h
+   include/liburing/io_uring.h
+   include/liburing/barrier.h
+   include/liburing/compat.h)
```
bikallem, line 21: TIL :+1:

lib/uring/include/discover.ml
```diff
@@ -2,49 +2,122 @@ module C = Configurator.V1
...
-    C.C_define.import c ~c_flags:["-D_GNU_SOURCE"]
-      ~includes:["fcntl.h"; "poll.h"; "sys/uio.h"; "linux/time_types.h"]
-      C.C_define.Type.[
-          "POLLIN", Int;
-          "POLLOUT", Int;
-          "POLLERR", Int;
-          "POLLHUP", Int;
+      let defs =
+        C.C_define.import c ~c_flags:["-D_GNU_SOURCE"; "-I"; Filename.concat (Sys.getcwd ()) "include"]
+          ~includes:["fcntl.h"; "poll.h"; "sys/uio.h"; "linux/time_types.h"; "liburing.h"]
```
bikallem, line 7: We can remove "linux/time_types.h" from here if we can successfully include "liburing.h". liburing defines `__kernel_timespce` if not defined already - https://github.com/axboe/liburing/blob/master/configure#L446-L453.

lib/uring/include/discover.ml
```diff
@@ -2,49 +2,122 @@ module C = Configurator.V1
...
+                    | "sizeof(struct __kernel_timespec)" -> "sizeof_kernel_timespec"
+                    | nm -> nm 
+                  in
+                  Printf.sprintf "let %s = 0x%x" (String.lowercase_ascii name) v
+                | _ -> assert false
+          )
+      in
+      let ops =
+        C.C_define.import c ~c_flags:["-D_GNU_SOURCE"; "-I"; Filename.concat (Sys.getcwd ()) "include"]
+          ~includes:["fcntl.h"; "poll.h"; "sys/uio.h"; "linux/time_types.h"; "liburing.h"]
```
bikallem, line 53: Same as above comment. `linux/time_types.h` is not needed if `liburing.h` can be included. Unless of course some other symbols defined in `linux/time_types.h` is needed.

lib/uring/include/discover.ml
talex5: OK, I've removed it now.

lib/uring/include/discover.ml
talex5: I've removed everything except `liburing.h` here.

PR #77, opened 2022-08-26
Remove is_dirty flag

talex5:
Instead of trying to track whether the ring is dirty, just ask liburing. Avoids this getting out of sync.

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -267,7 +267,6 @@ type 'a t = {
   data : 'a Heap.t;
   sketch : Sketch.t;
   queue_depth: int;
-  mutable dirty: bool; (* has outstanding requests that need to be submitted *)
 }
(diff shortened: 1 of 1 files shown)


haesbaert (approved):
The manual is pretty much contradicting itself, but I checked the code and it's fine.

The only difference is it needs a LOAD fence for the poll case, which is a compiler barrier in x86 and cheap enough in arm.

```DESCRIPTION
       The  io_uring_sq_ready(3)  function  retuns the number of unconsumed (if SQPOLL) or unsubmitted entries that exist in the SQ ring be‐
       longing to the ring param.

       Usage of this function only applies if the ring has been setup with IORING_SETUP_SQPOLL, where request submissions,  and  hence  con‐
       sumption from the SQ ring, happens through a polling thread.
```

PR #79, opened 2022-08-29
timeout: add support for completion_count in timeout call

bikallem:
This PR adds support for specifying 'completion (cqe) count' when issuing `timeout` sqe to uring. By specifying `completion_count`, timeout sqe returns `0` - success - when the specified number of `completion_count` cqes complete. 

The parameter is an optional type with default value of `0` to make it backwards compatible with earlier release.

I have also made a performance improvement to `set_timespec` c stub function by reducing the need to allocate `int64`.

`submit_timeout` c stub is also implemented in 2 c stubs since it has exceeded 5 parameters limit of OCaml stub functions.

### DIFF
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -73,12 +73,15 @@ type clock = Boottime | Realtime
 (** Represents Linux clocks. [Boottime] and [Realtime] represents OS clocks CLOCK_BOOTTIME
     and CLOCK_REALTIME respectively. *)
 
-val timeout: ?absolute:bool -> 'a t -> clock -> int64 -> 'a -> 'a job option
+val timeout: ?absolute:bool -> ?completion_count:int -> 'a t -> clock -> int64 -> 'a -> 'a job option
 (** [timeout t clock ns d] submits a timeout request to uring [t].
 
     [absolute] denotes how [clock] and [ns] relate to one another. Default value is [false]
 
-    [ns] is the timeout time in nanoseconds *)
+    [ns] is the timeout time in nanoseconds.
+
+    [completion_count] is the count of completion entries after which the timeout request times out.
+    Default value is '0'. *)
 
 module type FLAGS = sig
   type t = private int
--- a/lib/uring/uring.mli
+++ b/lib/uring/uring.mli
@@ -80,8 +80,12 @@ val timeout: ?absolute:bool -> ?completion_count:int -> 'a t -> clock -> int64 -
 
     [ns] is the timeout time in nanoseconds.
 
-    [completion_count] is the count of completion entries after which the timeout request times out.
-    Default value is '0'. *)
+    [completion_count] is the count of completion queue entries which must succeed - i.e. [0 = cqe->res] -
+    for the timeout sqe to also succeed, ie. [0 = timeout_cqe->res]. If the specified number of 
+    [completion_count] cqes are not reached within the given [ns] value, then the timeout request [cqe->res]
+    value is [-62] or [-ETIME]. [-62], [-ETIME] denotes that timeout [ns] was reached before uring could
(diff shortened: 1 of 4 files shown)


lib/uring/uring.mli
```diff
@@ -73,12 +73,15 @@ type clock = Boottime | Realtime
...
-val timeout: ?absolute:bool -> 'a t -> clock -> int64 -> 'a -> 'a job option
+val timeout: ?absolute:bool -> ?completion_count:int -> 'a t -> clock -> int64 -> 'a -> 'a job option
 (** [timeout t clock ns d] submits a timeout request to uring [t].
 
     [absolute] denotes how [clock] and [ns] relate to one another. Default value is [false]
 
-    [ns] is the timeout time in nanoseconds *)
+    [ns] is the timeout time in nanoseconds.
+
+    [completion_count] is the count of completion entries after which the timeout request times out.
```
talex5, line 83: I think `times out` is confusing here. The description in the PR made more sense to me:

> timeout request returns with 0 - success - when the specified number of completion_count cqes complete.

lib/uring/uring.mli
```diff
@@ -73,12 +73,15 @@ type clock = Boottime | Realtime
...
+    Default value is '0'. *)
```
talex5, line 84:
```suggestion
    Default value is '0', which disables this behaviour. *)
```

lib/uring/uring_stubs.c
```diff
@@ -151,17 +151,28 @@ ocaml_uring_sq_ready(value v_uring) {
...
 #define Val_boottime Val_int(0)
 
 value /* noalloc */
-ocaml_uring_submit_timeout(value v_uring, value v_id, value v_sketch_ptr, value v_clock, value v_absolute)
+ocaml_uring_submit_timeout_native(value v_uring,
+    value v_id,
+    value v_sketch_ptr,
+    value v_clock,
+    value v_absolute,
+    int completion_count)
```
talex5, line 175: The docs https://v2.ocaml.org/manual/intfc.html#s%3AC-cheaper-call say:
> Note: do not use the C int type in correspondence with `(int [@untagged])`. This is because they often differ in size.

But I don't think this should be an `int` anyway, as there's no untagged attribute (note that the MDX tests are run in byte-code mode).

tests/main.md
```diff
@@ -737,6 +737,32 @@ val timeout : int = -62
...
+# let ns1 = Int64.(mul 10L 1_000_000L);;
+val ns1 : int64 = 10000000L
+# let timeout_job = Uring.(timeout ~completion_count:1 t Boottime ns1 Timeout);;
+val timeout_job : job Uring.job option = Some <abstr>
+# Uring.submit t;;
+- : int = 2
+# let completed_job, res = consume t;;
+val completed_job : job = Noop
+val res : int = 0
+# Uring.get_cqe_nonblocking t;;
```
talex5, line 759: Not sure this is correct. I guess Linux will *probably* return both at once, and in this order, but I don't think it's guaranteed.

lib/uring/uring.mli
bikallem: I have updated the documentation.

lib/uring/uring.mli
bikallem: It doesn't disable the timeout request, just the cqe->res value is always -62/-ETIME.

lib/uring/uring_stubs.c
bikallem: Good catch. This is fixed now.

tests/main.md
bikallem: I have updated the tests to demonstrate that correlation between timeout return value and completion_count parameter.

PR #80, opened 2022-09-03
Add cscope.out and cscope.files to gitignore

haesbaert:


talex5:
It would probably be better to put these in your global git-ignores list. Otherwise everyone needs to submit files for every tool they use to every project they work on, which gets a bit excessive.

haesbaert:
no problemo

PR #81, opened 2022-09-03
Decouple Heap entries from ring size

haesbaert:
Before this change, we could only wait at most `queue_depth` events at a time,
we were wrongly tying the size of the CQ ring to the maximum number of syscalls
we could wait.

Example: I create a ring with queue_depth=1, in Uring this creates a CQ and SQ
ring, as well as a Heap with max storage of size 1. On the SQ side this means "I
can queue one element between each Uring.submit. On the CQ side this _should_
mean "I can wait _any_ number of events, but the kernel will only be able to put
one at the ring at a time, the others will be backlogged somewhere in the
kernel."

Imagine you have a ring(`queue_depth`) of 64, and 1000 TCP connections. Before,
we could only wait for events on 64 of these connections, we would literally
limit our submits by the size of the CQ ring. Now you can have a ring of 64, and
wait on """infinite""" TCP connections, but every time you wait on the ring, it
will have at most 64 entries (from _any_ of the 1000 connections).

This change grows the Heap when necessary, code is a bit primitive and basically
just resizes it, I made an attempt of Heap of Heaps but it became too convoluted
for very little gain.

We also need to work on respecting the return of Uring.submit, more
especifically:

```
EAGAIN - The kernel was unable to allocate memory for the request, or otherwise
ran out of resources to handle it. The application should wait for some
completions and try again.

EBUSY - The application is attempting to overcommit the number of requests it can
have pending. The application should wait for some completions and try again.
May occur if the application tries to queue more requests than we have room for
in the CQ ring, or if the application attempts to wait for more events without
having reaped the ones already present in the CQ ring.
 ```

I went over the tests that relied on the Heap not growing and changed it to
`alloc_no_growth`, but more eyes are necessary. I hate the name `no_growth` but
`can_fail` is also a lie since `growth` can fail if the Heap has already been
released or the allocation size blew up.

### DIFF
--- a/lib/uring/heap.ml
+++ b/lib/uring/heap.ml
@@ -28,11 +28,11 @@ type 'a entry =
 
 (* Free-list allocator *)
 type 'a t =
-  { data: 'a entry array
+  { mutable data: 'a entry array
   (* Pool of potentially-empty data slots. Invariant: an unfreed pointer [p]
      into this array is valid iff [free_tail_relation.(p) = slot_taken]. *)
   ; mutable free_head: ptr
-  ; free_tail_relation: ptr array
+  ; mutable free_tail_relation: ptr array
   (* A linked list of pointers to free slots, with [free_head] being the first
      element and [free_tail_relation] mapping each free slot to the next one.
      Each entry [x] signals a state of the corresponding [data.(x)] slot:
@@ -68,7 +68,52 @@ let create : type a. int -> a t =
 
 exception No_space
 
-let alloc t data ~extra_data =
+let in_use t = t.in_use
+
+let release t =
+  if t.in_use > 0 then invalid_arg "Heap still in use!"
+  else if t.in_use < 0 then invalid_arg "Heap already released!";
+  t.in_use <- -100;
+  t.free_head <- free_list_nil
+
+let is_released t = t.in_use < 0
+
+let grow t =
+  if is_released t then raise No_space;
+  let old_len = Array.length t.free_tail_relation in
+  if old_len = Sys.max_array_length then
+    raise No_space;
+  let new_len = min (max 64 (old_len * 2)) Sys.max_array_length in
+
+  (* Build new t.free_tail_relation, keep in sync with create() *)
+  let new_free_tail_relation =
+    Array.init new_len
+      (fun i ->
+         if i < old_len then
+           t.free_tail_relation.(i)
+         else succ i)
+  in
+  new_free_tail_relation.(new_len - 1) <- free_list_nil;
+
+  (* First element of enlarged array *)
+  let new_free_head = old_len in
+
+  (* Note: Keep in sync with create() *)
+  let new_data =
+    Array.init new_len
+      (fun i ->
+         if i < old_len then
+           t.data.(i)
+         else
+           Empty)
+  in
+
+  (* Commit *)
+  t.free_tail_relation <- new_free_tail_relation;
+  t.free_head <- new_free_head;
+  t.data <- new_data
+
+let alloc_no_growth t data ~extra_data =
   let ptr = t.free_head in
   if ptr = free_list_nil then raise No_space;
   let entry = Entry { data; extra_data; ptr } in
@@ -81,6 +126,12 @@ let alloc t data ~extra_data =
   t.in_use <- t.in_use + 1;
 
   entry
+
+let alloc t data ~extra_data =
+  try
+    alloc_no_growth t data ~extra_data
+  with
+    No_space -> grow t; alloc_no_growth t data ~extra_data
 
 let free t ptr =
   assert (ptr >= 0) (* [alloc] returns only valid pointers. *);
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: lib/uring/heap.ml [hunk 1]
```diff
@@ -68,7 +68,52 @@ let create : type a. int -> a t =
...
+
+let release t =
+  if t.in_use > 0 then invalid_arg "Heap still in use!"
+  else if t.in_use < 0 then invalid_arg "Heap already released!";
+  t.in_use <- -100;
+  t.free_head <- free_list_nil
+
+let is_released t = t.in_use < 0
+
+let grow t =
```
talex5, line 81: I guess there's a precondition here that the array is currently full? Might be worth stating it explicitly.

lib/uring/heap.ml
```diff
@@ -68,7 +68,52 @@ let create : type a. int -> a t =
...
+  if is_released t then raise No_space;
+  let old_len = Array.length t.free_tail_relation in
+  if old_len = Sys.max_array_length then
+    raise No_space;
+  let new_len = min (max 64 (old_len * 2)) Sys.max_array_length in
+
```
talex5, line 87: Minor style note: I don't think we put blank lines in OCaml functions normally.

talex5 (commented):
Thanks - this needed fixing!

> I went over the tests that relied on the Heap not growing and changed it to `alloc_no_growth`, but more eyes are necessary.

There's no point exposing a non-growing heap that no-one uses just so we can unit test it! And it would simplify the code to remove this.

haesbaert:
> Thanks - this needed fixing!
> 
> > I went over the tests that relied on the Heap not growing and changed it to `alloc_no_growth`, but more eyes are necessary.
> 
> There's no point exposing a non-growing heap that no-one uses just so we can unit test it! And it would simplify the code to remove this.

I'd prefer to do this in a subsequent PR, I was also considering that this code is generic enough to be copied elsewhere where it might have use, but at any rate I'd prune all the no_growth bits later, if that's ok.

talex5:
> and also didn't want to break the tests existing invariants

If a test breaks because it's checking the help can't grow, just delete it. It's really confusing having tests ensuring the code does something we don't care about. These tests have already made this PR messier (and maybe slower) that it should be, and it will get worse next time someone tries to refactor it and tries to guess what the tests are for.

Anyway, looks fine apart from that!

bikallem:
>Imagine you have a ring(queue_depth) of 64, and 1000 TCP connections.

Does this mean that so far, if we try to create 1000 TCP connections, we were only able to create 64? Since I think creation of sockets is also a SQE, no??

haesbaert:
> 

You can think of this: if there was 64 existing connections, and all of them were just waiting for data that never comes, every operation that needed a SQE would hang until one of the the 64 connections got at least one packet.

haesbaert:
> > and also didn't want to break the tests existing invariants
> 
> If a test breaks because it's checking the heap can't grow, just delete it. It's really confusing having tests ensuring the code does something we don't care about. These tests have already made this PR messier (and maybe slower) that it should be, and it will get worse next time someone tries to refactor it and tries to guess what the tests are for.
> 
> Anyway, looks fine apart from that!

Ack, I'll remove it then, had a bad spell the last two days and couldn't get much done, will work on it tomorrow

lib/uring/uring.ml
```diff
@@ -344,19 +344,17 @@ let exit t =
...
-    let has_space = fn ptr in
-    if has_space then (
-      Some entry
-    ) else (
-      ignore (Heap.free t.data ptr : a);
-      None
-    )
+ let entry = try Heap.alloc t.data datum ~extra_data
+   with exc -> check t; raise exc
+ in
```
talex5, line 349: Does this code really need to be reformatted? It seemed OK as it was (the new code is only shorter because the comment got deleted).
```suggestion
   match Heap.alloc t.data datum ~extra_data with
  | exception (Invalid_argument _ as ex) -> check t; raise ex
  | entry ->
```

talex5 (commented):
> the benchmarks only end up growing on the end of bench_noop

Might be due to the extra asserts. I think they can be moved out of the fast path.

Apart from that, OK to merge.

lib/uring/heap.ml
```diff
@@ -66,20 +66,65 @@ let create : type a. int -> a t =
...
+         else
+           Empty)
+  in
+  (* Commit *)
+  t.free_tail_relation <- new_free_tail_relation;
+  t.free_head <- new_free_head;
+  t.data <- new_data
 
 let alloc t data ~extra_data =
+  maybe_already_released t;
```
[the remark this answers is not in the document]
haesbaert: ah good point

lib/uring/heap.ml
```diff
@@ -66,20 +66,65 @@ let create : type a. int -> a t =
...
+  if t.free_head = free_list_nil then grow t;
   let ptr = t.free_head in
-  if ptr = free_list_nil then raise No_space;
+  assert (ptr <> free_list_nil);
```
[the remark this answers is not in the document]
haesbaert: this was more for peace of mind and coding defensively, I'll remove

lib/uring/uring.ml
haesbaert: I was shuffling things around and this snuck in.

haesbaert:
> 

I also ran benchmarks without the asserts and extra raises and whatnot, but there was too much noise too see anything

PR #82, opened 2022-11-22
update to liburing-2.3

avsm:
This adds support for aarch64/musl and zerocopy transmit.

A potential breaking change is the return value fix to io_uring_submit_and_wait_timeout, but our usecases aren't affected.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,23 @@
+# Maintainer notes
+
+## liburing updates
+
+`vendor/liburing/` contains the upstream version of liburing. Updating it
+is normally done using `git subtree`.
+
+To do an update, first fetch the remote references from the liburing repo:
+
+```
+git remote add liburing https://github.com/axboe/liburing.git
+git fetch
+```
+
+Then run a subtree pull, and substitute `2.3` with the right version:
+
+```
+git subtree pull --prefix vendor/liburing https://github.com/axboe/liburing.git liburing-2.3 --squash
+```
+
+Then modify the build rules over at `lib/uring/dune` to pick up the
+right version of the shared library that's built.
+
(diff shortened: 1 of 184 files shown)


talex5:
This commit looks different to the previous subtree updates (which appear as merges, e.g. #56). I think the command I normally use is:
```
git subtree pull --prefix vendor/liburing https://github.com/axboe/liburing.git $BRANCH --squash
```

avsm:
I always find subtree more hassle than it's worth... in this case, the `subtree pull` doesn't work for me:

```
$ git subtree pull --prefix vendor/liburing https://github.com/axboe/liburing.git liburing-2.3 --squash
From https://github.com/axboe/liburing
 * tag                 liburing-2.3 -> FETCH_HEAD
fatal: ambiguous argument '994a08df855bdcac7bf1e3e866a286f62b90b45d^{commit}': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
could not rev-parse split hash 994a08df855bdcac7bf1e3e866a286f62b90b45d from commit c9a690edb37d494ee77051707a8b14b685df4e82
```

...but adding liburing as a remote, fetching, and then doing the subtree pull works.  Which is weird, because I thought the whole point of subtree was to avoid mixing up the upstream repository commits with the local one!

HACKING.md
```diff
@@ -0,0 +1,23 @@
...
+## liburing updates
+
+`vendor/liburing/` contains the upstream version of liburing. Updating it
+is normally done using `git subtree`.
+
+To do an update, first fetch the remote references from the liburing repo:
+
+```
+git remote add liburing https://github.com/axboe/liburing.git
+git fetch
```
avsm, line 12: I'm not sure this is the right way to do this...

talex5 (approved):

PR #86, opened 2023-02-15
Use `caml_enter_blocking_section` when calling `io_uring_submit`

TheLortex:
`io_uring_submit` sends asynchronous work to the kernel but it still has to take a copy the data, which may take some time. In my experiments (read operations of 256MB) the system call can take up to 200ms, which is not good when domains need to synchronize.  

Using `caml_enter_blocking_section` notifies the runtime that we're leaving the OCaml world, similarly to https://github.com/ocaml-multicore/ocaml-uring/pull/31.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 ## dev
 
-- Use `caml_enter_blocking_section` when calling `io_uring_submit`  (@TheLortex #84).
+- Use `caml_enter_blocking_section` when calling `io_uring_submit`  (@TheLortex #86).
 
 ## v0.5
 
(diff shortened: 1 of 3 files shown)


talex5 (commented):
The function can't be `noalloc` if it releases the lock: https://v2.ocaml.org/manual/intfc.html#ss:c-direct-call

TheLortex:
Ah indeed, thanks for noticing. I'll update the PR.

avsm:
Looks good to me. I think there are some situations where we might want the faster original version. In particular, if the caller knows they are using fixed buffers (which don't need to be copied) then the syscall should be pretty fast. However in the general case we should drop the lock, so this PR is a safer default.

talex5 (approved):
Thanks!

Issue #88, opened 2023-05-14
SIGSEGV on connect


edwintorok:
Here is a minimized testcase that causes `uring` to segfault:
```
let rec consume uring =
  match Uring.wait ~timeout:1.0 uring with
  | None ->
      ()
  | Some _ ->
      consume uring

let sockets =
  Array.init 500 (fun _ ->
      let s = Unix.socket Unix.PF_INET Unix.SOCK_STREAM 0 in
      (s, Unix.ADDR_INET (Unix.inet_addr_of_string "127.0.0.1", 631)) )

let uring = Uring.create ~polling_timeout:1000 ~queue_depth:128 ()

let () =
  let () =
    sockets
    |> Array.iter
       @@ fun (sock, dest) ->
       let rec loop () =
         match Uring.connect uring sock dest 5 with
         | Some _ ->
             ()
         | None ->
             let (_ : 'a Uring.completion_option) =
               Uring.wait ~timeout:1.0 uring
             in
             loop ()
       in
       loop ()
  in
  while Uring.active_ops uring > 0 do
    consume uring
  done
```

Compile and run like this (assuming you have something listening on port 631 localhost, if not pick another port...):
```
$ ocamlfind ocamlopt -package unix,uring conn.ml -linkpkg -o conn
$ ./conn
[1]    112639 segmentation fault (core dumped)  ./conn
```

Software and hardware environment:
```
$ ocamlopt --version
5.0.0
$ uname -a
Linux fedora38.fedora 6.2.14-300.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Mon May  1 00:55:28 UTC 2023 x86_64 GNU/Linux
$ inxi
CPU: 16-core AMD Ryzen 9 7950X (-MT MCP-) speed/min/max: 3131/3000/5880 MHz
Kernel: 6.2.14-300.fc38.x86_64 x86_64 Up: 10h 4m Mem: 5939.9/63423.8 MiB
(9.4%) Storage: 24.99 TiB (4.0% used) Procs: 536 Shell: Zsh inxi: 3.3.26
```

What I tried:
* changing the number of sockets from 500 to 50: no crash
* removing the '~timeout' flag from `Uring.wait`: no crash, but stuck
* removing the `~polling_timeout` flag: no crash, but stuck
* using the debug runtime: didn't provide more info about the crash

`coredumpctl gdb`:
```
Core was generated by `_build/default/loadtest_uring.exe http://localhost:631'.
[...]

PR #89, opened 2023-05-14
Fix SIGSEGV on `Uring.wait`

edwintorok:
Fixes https://github.com/ocaml-multicore/ocaml-uring/issues/88

### DIFF
--- a/lib/uring/region.ml
+++ b/lib/uring/region.ml
@@ -5,7 +5,6 @@
 type t = {
   buf: Cstruct.buffer;
   block_size: int;
-  slots: int;
   freelist: int Queue.t;
 }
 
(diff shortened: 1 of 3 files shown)


talex5:
Looks like the behaviour of `io_uring_submit_and_wait_timeout` changed (see https://github.com/axboe/liburing/issues/684). Previously it returned `ETIME` for a timeout, but now if it managed to submit something then it returns the number of items submitted instead. Therefore, you can now get success with no CQE.

I've added a test to your branch to check the fix.

talex5 (approved):
Thanks!

avsm:
Good catch, thanks @edwintorok!

PR #96, opened 2023-08-11
Add type annotations to tests

patricoferris:
It would seem between OCaml 5 and 5.1, the type variable part of polymorphic variants is now shown in the top-level which means between 5 and 5.1 the tests are failing and we can't promote to either without breaking things. To avoid this issue I've annotated all of the `Uring.t`s

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ To start, we'll open a file using `Uring.openat2`, which works much like the reg
       ~resolve:Uring.Resolve.beneath
       "test.log"
       `Open_log;;
-val open_file : _[> `Open_log ] Uring.job option = Some <abstr>
+val open_file : ([> `Open_log ] as '_weak2) Uring.job option = Some <abstr>
 ```
 
 `submit` returns `None` if the submission queue is full.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,8 +20,8 @@ To use the library directly, you need to use the `uring` ocamlfind library:
 Call `Uring.create` to initialise a ring:
 
 ```ocaml
-# let uring = Uring.create ~queue_depth:10 ();;
-val uring : '_weak1 Uring.t = <abstr>
+# let uring : [ `Open_log | `Close_log | `Write_all ] Uring.t = Uring.create ~queue_depth:10 ();;
+val uring : [ `Close_log | `Open_log | `Write_all ] Uring.t = <abstr>
 ```
 
 The `'_weak1` is the type of user data attached to requests, which can be whatever you want.
```
talex5, line 27: For the README, it might be better to restrict it to OCaml 5.1 in the MDX. Then we can still show it doing inference.

Alternatively, the text needs changing so it no longer mentions `'_weak1`.

talex5 (commented):
Thanks - just needs a fix for the README.

README.md
patricoferris: Ah missed the comment, thanks! Forced pushed to use an `enabled_if` in the `dune` file 👍

PR #97, opened 2023-08-15
Statx opt

patricoferris:
A follow on PR to #95 after @avsm's review. Thought I'd open the PR early to get some eyes on it but I would like to double-check I didn't make (more) copy-paste mistakes, mess up the noalloc functions etc. 

Instead of having individual C functions we could perhaps have an "int64_field" function and choose the right field based on a flag we send in from OCaml which might reduce the amount of code... but then internally there's a big switch statement so I wasn't sure if that was any better. 

Also I haven't added the `STATX_DIOALIGN` flag as it is Linux 6.1 ?

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -144,6 +144,7 @@ module Statx = struct
   external uid : t -> (int64 [@unboxed]) = "ocaml_uring_statx_uid_bytes" "ocaml_uring_statx_uid_native" [@@noalloc]
   external gid : t -> (int64 [@unboxed]) = "ocaml_uring_statx_gid_bytes" "ocaml_uring_statx_gid_native" [@@noalloc]
   external ino : t -> (int64 [@unboxed]) = "ocaml_uring_statx_ino_bytes" "ocaml_uring_statx_ino_native" [@@noalloc]
+  external size : t -> (int64 [@unboxed]) = "ocaml_uring_statx_size_bytes" "ocaml_uring_statx_size_native" [@@noalloc]
   external blocks : t -> (int64 [@unboxed]) = "ocaml_uring_statx_blocks_bytes" "ocaml_uring_statx_blocks_native" [@@noalloc]
   external attributes_mask : t -> (int64 [@unboxed]) = "ocaml_uring_statx_attributes_mask_bytes" "ocaml_uring_statx_attributes_mask_native" [@@noalloc]
   external rdev : t -> (int64 [@unboxed]) = "ocaml_uring_statx_rdev_bytes" "ocaml_uring_statx_rdev_native" [@@noalloc]
@@ -167,7 +168,6 @@ module Statx = struct
   external perm : t -> (int [@untagged]) = "ocaml_uring_statx_perm_bytes" "ocaml_uring_statx_perm_native" [@@noalloc]
 
   external kind : t -> kind = "ocaml_uring_statx_kind"
-  external size : t -> Optint.Int63.t = "ocaml_uring_statx_size"
 end
 
 module Sockaddr = struct
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -100,6 +100,18 @@ module Statx = struct
     | `Socket
   ]
 
+  let pp_kind f k =
+    Fmt.pf f "%s"
+      (match k with
+       |`Unknown -> "unknown"
+       |`Fifo -> "fifo"
+       |`Character_special -> "character special file"
+       |`Directory -> "directory"
+       |`Block_device -> "block device"
+       |`Regular_file -> "regular file"
+       |`Symbolic_link -> "symbolic link"
+       |`Socket -> "socket")
+
   external create : unit -> t = "ocaml_uring_make_statx"
 
   module Flags = struct
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -153,10 +153,15 @@ module Statx = struct
   external dio_mem_align : t -> (int64 [@unboxed]) = "ocaml_uring_statx_dio_mem_align_bytes" "ocaml_uring_statx_dio_mem_align_native" [@@noalloc]
   external dio_offset_align : t -> (int64 [@unboxed]) = "ocaml_uring_statx_dio_offset_align_bytes" "ocaml_uring_statx_dio_offset_align_native" [@@noalloc]
 
-  external atime : t -> (float [@unboxed]) = "ocaml_uring_statx_atime_bytes" "ocaml_uring_statx_atime_native" [@@noalloc]
-  external btime : t -> (float [@unboxed]) = "ocaml_uring_statx_btime_bytes" "ocaml_uring_statx_btime_native" [@@noalloc]
-  external ctime : t -> (float [@unboxed]) = "ocaml_uring_statx_ctime_bytes" "ocaml_uring_statx_ctime_native" [@@noalloc]
-  external mtime : t -> (float [@unboxed]) = "ocaml_uring_statx_mtime_bytes" "ocaml_uring_statx_mtime_native" [@@noalloc]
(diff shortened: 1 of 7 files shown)


patricoferris:
For clarity on the int64 comment something like: 

```ocaml
external int64_field : internal -> (int [@untagged]) -> (int64 [@unboxed]) = "..." "..." [@@noalloc]

let blksize t = int64_field t Mask.blksize
```

Which would then switch on the field 

```c
int64_t
ocaml_uring_statx_int64_field(value v_statx, value v_mask) {
  struct statx *s = Statx_val(v_statx);
  switch (v_mask) {
     case STATX_BLKSIZE: ...
  }
}
```

But then it might have to raise an exception (so can't be noalloc I think) or have a default return value which I wasn't sure what that should be.

avsm:
Looks like the right direction! I'd also:
- remove `Statx.t` -- I don't think it's a good idea to encourage reading statx field values that aren't explicitly filled in by the mask bit for that field being set, and this type encourages the user of the interface to do that.  The man page isn't clear on if those fields are changed or not after the invocation to statx, but right now they won't be zeroed out between uses of `Statx.internal`. 
- rename `Statx.internal` to `Statx.t` or `Statx.buf`
- re: the extra fields that are 5.8 or 6.1+, we could just set them to be detected to 0 if the relevant define isn't found. This is safe since the return mask would not be filled in even if the flag is specified in the statx.
- Could also define the following for `file_attributes`
```
       STATX_ATTR_COMPRESSED
       STATX_ATTR_IMMUTABLE
       STATX_ATTR_APPEND
       STATX_ATTR_NODUMP
       STATX_ATTR_ENCRYPTED
       STATX_ATTR_VERITY
       STATX_ATTR_DAX
```

avsm:
(The individual C accessors seem fine to me)

lib/uring/uring_stubs.c
```diff
@@ -524,6 +497,255 @@ ocaml_uring_submit_statx_byte(value* values, int argc) {
 			  values[6]);
 }
 
+// Non-allocating accessors for statx
+// Int64
+int64_t
+ocaml_uring_statx_blksize_native(value v_statx) {
```
talex5, line 503: Might be worth defining a macro to make these functions automatically. e.g. something like:
```c
STATX_GETTER(blksize, int64_t, caml_copy_int64)
STATX_GETTER(attributes, int64_t, caml_copy_int64)
...
```

lib/uring/uring.ml
```diff
@@ -138,6 +125,32 @@ module Statx = struct
...
+  external ino : t -> (int64 [@unboxed]) = "ocaml_uring_statx_ino_bytes" "ocaml_uring_statx_ino_native" [@@noalloc]
+  external blocks : t -> (int64 [@unboxed]) = "ocaml_uring_statx_blocks_bytes" "ocaml_uring_statx_blocks_native" [@@noalloc]
+  external attributes_mask : t -> (int64 [@unboxed]) = "ocaml_uring_statx_attributes_mask_bytes" "ocaml_uring_statx_attributes_mask_native" [@@noalloc]
+  external rdev : t -> (int64 [@unboxed]) = "ocaml_uring_statx_rdev_bytes" "ocaml_uring_statx_rdev_native" [@@noalloc]
+  external dev : t -> (int64 [@unboxed]) = "ocaml_uring_statx_dev_bytes" "ocaml_uring_statx_dev_native" [@@noalloc]
+  external mask : t -> (int64 [@unboxed]) = "ocaml_uring_statx_mask_bytes" "ocaml_uring_statx_mask_native" [@@noalloc]
+  external mnt_id : t -> (int64 [@unboxed]) = "ocaml_uring_statx_mnt_id_bytes" "ocaml_uring_statx_mnt_id_native" [@@noalloc]
+  external dio_mem_align : t -> (int64 [@unboxed]) = "ocaml_uring_statx_dio_mem_align_bytes" "ocaml_uring_statx_dio_mem_align_native" [@@noalloc]
+  external dio_offset_align : t -> (int64 [@unboxed]) = "ocaml_uring_statx_dio_mem_align_bytes" "ocaml_uring_statx_dio_mem_align_native" [@@noalloc]
```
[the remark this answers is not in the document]
patricoferris: oops thank you!

lib/uring/uring_stubs.c
patricoferris: Good idea, did that where feasible here https://github.com/ocaml-multicore/ocaml-uring/pull/97/commits/b12ae5ec19c9a90e4d31cdebc76ac6363f485414

lib/uring/uring_stubs.c
```diff
@@ -524,6 +497,136 @@ ocaml_uring_submit_statx_byte(value* values, int argc) {
...
+STATX_GETTER(blksize, int64_t, caml_copy_int64);
+STATX_GETTER(attributes, int64_t, caml_copy_int64);
+STATX_GETTER(nlink, int64_t, caml_copy_int64);
+STATX_GETTER(uid, int64_t, caml_copy_int64);
+STATX_GETTER(gid, int64_t, caml_copy_int64);
+STATX_GETTER(ino, int64_t, caml_copy_int64);
+STATX_GETTER(blocks, int64_t, caml_copy_int64);
+STATX_GETTER(attributes_mask, int64_t, caml_copy_int64);
+STATX_GETTER(mask, int64_t, caml_copy_int64);
+
```
talex5, line 520: Minor: it would be easier to read with the field name at the end, so the bits that don't change line up vertically:
```suggestion
STATX_GETTER(int64_t, caml_copy_int64, blksize);
STATX_GETTER(int64_t, caml_copy_int64, attributes);
STATX_GETTER(int64_t, caml_copy_int64, nlink);
STATX_GETTER(int64_t, caml_copy_int64, uid);
STATX_GETTER(int64_t, caml_copy_int64, gid);
STATX_GETTER(int64_t, caml_copy_int64, ino);
STATX_GETTER(int64_t, caml_copy_int64, blocks);
STATX_GETTER(int64_t, caml_copy_int64, attributes_mask);
STATX_GETTER(int64_t, caml_copy_int64, mask);
```

talex5 (approved):

avsm:
Actually only observing the different time resolution results on directories/symlinks, not regular files.

```
$ stat /lib
  File: /lib -> usr/lib
  Size: 7               Blocks: 0          IO Block: 4096   symbolic link
Device: 8,3     Inode: 15          Links: 1
Access: (0777/lrwxrwxrwx)  Uid: (    0/    root)   Gid: (    0/    root)
Access: 2023-08-15 22:17:01.209919458 +0000
Modify: 2023-01-24 04:23:40.355852338 +0000
Change: 2023-02-11 21:42:14.626724408 +0000
 Birth: 2023-02-11 21:42:14.626724408 +0000
$ dune exec -- tests/urstat.exe /lib
  File: /lib -> usr/lib               
  Size: 7               Blocks: 0       IO Block: 4096   symbolic link
Device: 2051    Inode: 15       Links: 1
Access: (0777/TODO)     Uid: (0/TODO)   Gid: (0/TODO)
Access: 2023-08-15 22:17:01.209919453 +0000
Modify: 2023-01-24 04:23:40.355852365 +0000
Change: 2023-02-11 21:42:14.626724482 +0000
 Birth: 2023-02-11 21:42:14.626724482 +0000
```

talex5:
> The time precision seems slightly off; are we rounding off tv_nsec? Should probably propagate it through rather than pack it into one 64-bit float, to avoid losing precision.

I've pushed a commit splitting them out. Indeed you can't represent the time exactly as a float (OCaml stdlib and Python both have the same limitation, though):

```
utop # 1662991022. +. (521759700. /. 1e9);;
- : float = 1662991022.52175975

utop # Float.next_after 1662991022.52175975 0.0;;
- : float = 1662991022.52175951
```

avsm:
Looks good, the urstat and stat outputs are exactly the same now. I pushed d0cd9d9564246e903d574ef21f2d383c763cce7a to make the size parameter a full int64 return now as well, so this PR is good to merge for me.

PR #100, opened 2023-08-19
add more ocamldoc

avsm:


lib/uring/uring.mli
```diff
@@ -91,42 +97,148 @@ module type FLAGS = sig
...
+  (** [sync] ensures that write operations on the file complete
+      according to the requirements of synchronised IO file integrity
+      completion. *) (* TODO bind fsync *)
+
   val tmpfile : t
+  (** [tmpfile] creates an anonymous temporary regular file. The pathname
+      must be a directory, within which an unnamed inode will be created.
+      If [tmpfile] is specified without {!excl}, then a subsequent
+      linkat call can move it permanently into the filesystem. *)
+  (* TODO bind symlinkat and linkat *)
```
avsm, line 178: I also can't find any hint of readlinkat support in io_uring upstream, for some reason.

lib/uring/uring.mli
```diff
@@ -91,42 +97,148 @@ module type FLAGS = sig
 
   val mem : t -> t -> bool
   (** [mem x flags] is [true] iff [x] is a subset of [flags]. *)
+
+  val empty : t
+  (** The empty set of flags *)
```
avsm, line 102: Interface change as well: I think it should always be possible to pass in empty flags to the various calls here.

lib/uring/uring.mli
```diff
@@ -91,42 +97,148 @@ module type FLAGS = sig
...
+  (** [path] will obtain a fd that can only be used to either indicate
+      a location in a filesystem tree, or perform operations at the fd
+      level. The file is not opened, and so any IO operations on the file
+      will fail. [path] is only used with {!cloexec}, {!directory} and
+      {!nofollow}, and any other flags will be ignored. *)
+
   val sync : t
+  (** [sync] ensures that write operations on the file complete
+      according to the requirements of synchronised IO file integrity
+      completion. *) (* TODO bind fsync *)
```
avsm, line 171: added in #103

lib/uring/uring.mli
```diff
@@ -14,7 +14,12 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-(** Io_uring interface. *)
+(** Io_uring is an asynchronous I/O API for Linux that uses ring buffers
+    shared between the Linux kernel and userspace to provide an efficient
+    mechanism to batch requests that can be handled asynchronously and in
+    parallel.  This module provide an OCaml interface to io_uring that
```
talex5, line 20:
```suggestion
    parallel.  This module provides an OCaml interface to io_uring that
```

lib/uring/uring.mli
```diff
@@ -91,42 +97,148 @@ module type FLAGS = sig
...
   val excl : t
+  (** [excl] is used alongside {!creat} to ensure that the file
+      is created as a result of the {!openat2} call, and otherwise
+      fails with a {!Unix.EEXIST} exception. The only exception where
+      [excl] can be used without {!creat} is when attempting to open
+      block devices. If the block device is otherwise mounted, then
+      the open will fail with {!Unix.EBUSY}. *)
+
   val largefile : t
+  (* TODO this can be hidden, since we should always set _FILE_OFFSET_BITS=1
```
talex5, line 143: Good idea.

lib/uring/uring.mli
```diff
@@ -278,36 +432,66 @@ module Statx : sig
   end
 
   module Mask : sig
+    (** The mask flags are used to tell the kernel which fields the {!statx} invocation
+        is interested in. You may wish to use {! Mask.check} on the returned {!Statx.t} to
+        verify the field has actually been filled in with a sensible value first.
+        In general, the kernel never refused values specified in the mask, but may choose
+        to not set the mask in the returned buffer from {!statx}. *)
```
talex5, line 439: Where is this from? statx(2) says:
> It should be noted that the kernel may return fields that weren't requested and may fail to return fields that were  requested

talex5 (commented):
Thanks - this looks pretty useful!

lib/uring/uring.mli
avsm: That quote is compatible with the ocamldoc above. From https://man7.org/linux/man-pages/man2/statx.2.html

       Note that, in general, the kernel does not reject values in mask
       other than the above.  (For an exception, see EINVAL in errors.)
       Instead, it simply informs the caller which values are supported
       by this kernel and filesystem via the statx.stx_mask field.
       Therefore, do not simply set mask to UINT_MAX (all bits set), as
       one or more bits may, in the future, be used to specify an
       extension to the buffer.

avsm:
mergeable now, but ocaml-ci is down...

lib/uring/uring.mli
```diff
@@ -93,40 +99,138 @@ module type FLAGS = sig
...
   val append : t
+  (** [append] repositions the file offset to the end of the file before
+      every write. This should be used with caution with io_uring.
+      @see <https://github.com/axboe/liburing/issues/32#issuecomment-1313220682> GitHub axboe/liburing#32 *)
+
   val cloexec : t
+  (** [cloexec] enables the close-on-exec flag for the new fd. *)
+
   val creat : t
+  (** [create] implies that if the pathname does not exist, it is
```
patricoferris, line 115:
```suggestion
  (** [creat] implies that if the pathname does not exist, it is
```

lib/uring/uring.mli
```diff
@@ -93,40 +99,138 @@ module type FLAGS = sig
...
+      completion. *)
+
   val tmpfile : t
+  (** [tmpfile] creates an anonymous temporary regular file. The pathname
+      must be a directory, within which an unnamed inode will be created.
+      If [tmpfile] is specified without {!excl}, then a subsequent
+      linkat call can move it permanently into the filesystem. *)
+
   val trunc : t
+  (** [trunc] will set the file size of 0 if the file already exists
```
patricoferris, line 173:
```suggestion
  (** [trunc] will set the file size to 0 if the file already exists
```

lib/uring/uring.mli
```diff
@@ -241,34 +345,76 @@ module Statx : sig
...
+
     val no_automount : t
+    (** [no_automount] signals that statx should not automount the basename
+        component of the path if it is an automount point.
+        This can be used in tools that scan directories to prevent
+        mass-automounting of a directory of automount points. *)
+
     val symlink_nofollow : t
+    (** [symlink_nofollow] signals that if the path is a symbolic link,
+        then return return information about the link itself. *)
```
patricoferris, line 371:
```suggestion
        then return information about the link itself. *)
```

patricoferris (approved):
Much more user friendly! I think the CI issues may be related to https://github.com/ocurrent/solver-service/issues/74

lib/uring/uring.mli
avsm: fixed in f9769b8d1cfe14a619182467dec21aef1965b2f5

lib/uring/uring.mli
avsm: fixed in f9769b8d1cfe14a619182467dec21aef1965b2f5

lib/uring/uring.mli
talex5: OK, I guess the meaning of "reject/refuse" here was confusing me.

I read the man-page as saying that you won't get `EINVAL` for specifying an unknown mask bit; instead it will just ignore the unknown bits for forwards-compatibility.

But `the kernel never refuse[s] values specified in the mask` could also be interpreted as saying that the kernel will always return the values you ask for.

PR #101, opened 2023-08-21
[Experiment] Cstruct to Bstruct

patricoferris:
This is an experiment to switch away from Cstruct to Bstruct (bytes-backed Cstruct). The main issue is that the bytes are in the OCaml heap but we have to ensure they don't move. I started playing around with an allocator (`Uring.Slab`) which allocates a big bytes area upfront and the carves it up as the user requests slices (returned as `Bstruct.t`). For this to work properly the `Uring.Slab` implementation needs a few more tricky features:

 - The bytes array probably needs to be more dynamic so it can create other big bytes areas when there's a lot of requests for slices.
 - Slices that are finished with, should be garbage-collected which probably means having some reference counting.

Opening this PR in case other people want to have a play around. Note this will only work with OCaml 5+ and is related to https://github.com/ocaml-multicore/eio/issues/140

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,12 +100,16 @@ let rec write_all fd = function
     let result, data = wait_with_retry uring in
     assert (data = `Write_all);  (* There aren't any other requests pending *)
     assert (result > 0);         (* Check for error return *)
-    let bufs = Uring.Bytes.shiftv bufs result in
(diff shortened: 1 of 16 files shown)


patricoferris:
Closing in favour of #153

PR #104, opened 2023-09-12
Add eventfd support

LucaSeri:
Registering an eventfd to the io_uring is useful for interfacing with systems that already use polling.


talex5 (approved):
Looks good - thanks!

Issue #112, opened 2024-04-24
Fails to build with musl 1.2.5


alyssais:
musl 1.2.5 newly provides some statx support in <sys/stat.h>, but it does not define the `STATX_ATTR_*` constants, causing this build failure:

```
run: gcc -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC  -D_GNU_SOURCE -I /build/uring-0.8/_build/default/lib/uring/include -I /nix/store/yspc3sc7nkgspq2l0larmf46flhk1r4n-ocaml-5.0.0/lib/ocaml -o /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.o -c /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c -lm -lpthread
-> process exited with code 1
-> stdout:
-> stderr:
 | /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c:198:14: error: ‘STATX_ATTR_COMPRESSED’ undeclared here (not in a function)
 |   198 |   DUNE_SIGN((STATX_ATTR_COMPRESSED)),
 |       |              ^~~~~~~~~~~~~~~~~~~~~
 | /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c:33:24: note: in definition of macro ‘DUNE_SIGN’
 |    33 | #define DUNE_SIGN(x) ((x >= 0)? '0': '-')
 |       |                        ^
 | /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c:206:14: error: ‘STATX_ATTR_IMMUTABLE’ undeclared here (not in a function)
 |   206 |   DUNE_SIGN((STATX_ATTR_IMMUTABLE)),
 |       |              ^~~~~~~~~~~~~~~~~~~~
 | /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c:33:24: note: in definition of macro ‘DUNE_SIGN’
 |    33 | #define DUNE_SIGN(x) ((x >= 0)? '0': '-')
 |       |                        ^
 | /build/build_458c2a_dune/ocaml-configurator434ca2/c-test-4/test.c:214:14: error: ‘STATX_ATTR_APPEND’ undeclared here (not in a function)
 |   214 |   DUNE_SIGN((STATX_ATTR_APPEND)),
 |       |              ^~~~~~~~~~~~~~~~~
[...]

PR #114, opened 2024-04-28
Fix statx constant fallback values, and fix with musl 1.2.5

alyssais:
Fixes https://github.com/ocaml-multicore/ocaml-uring/issues/112.

### DIFF
--- a/lib/uring/include/discover.ml
+++ b/lib/uring/include/discover.ml
@@ -116,6 +116,11 @@ let () =
             "STATX_MNT_ID", Int, 0x00001000;
             "STATX_DIOALIGN", Int, 0x00002000;
             (* File Attributes *)
+            "STATX_ATTR_COMPRESSED", Int, 0x00000004;
+            "STATX_ATTR_IMMUTABLE", Int, 0x00000010;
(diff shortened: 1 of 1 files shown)


alyssais:
Hmm, CI gets errors because some of them are unused.  I guess I should just remove those ones?

alyssais:
Although… shouldn't they be used in `Attr`?

talex5:
It's because you're generating some of them twice. Looking at _build/default/lib/uring/config.ml:
```
  module Attr = struct
    let dax = 0x200000
    let verity = 0x100000
    let mount_root = 0x2000
    let automount = 0x1000
    let encrypted = 0x800
    let nodump = 0x40                       <<<
    let append = 0x20
    let immutable = 0x10
    let compressed = 0x4
    let encrypted = 0x800
    let nodump = 0x40                      <<<
    let append = 0x20
    let immutable = 0x10
    let compressed = 0x4
  end
```

alyssais:
Ah, of course.  Let's try this.

talex5 (approved):
Thanks, that's great!

PR #127, opened 2025-09-28
Fix mkdir test

talex5:
According to `man 7 inode`:
> POSIX refers to the stat.st_mode bits corresponding to the mask S_IFMT
> (see below) as the file type, the 12 bits  corresponding to the mask
> 07777 as the file mode bits and the least significant 9 bits (0777) as
> the file permission bits.

OCaml defines `st_perm` as:
> The type of file access rights, e.g. 0o640 is read and write for user,
> read for group, none for others

However, OCaml uses `buf->st_mode & 07777` and so includes three other bits. We need to mask them out for the test (see failure in https://github.com/ocaml/opam-repository/pull/28604).

/cc @patricoferris


avsm:
Not sure why I didn’t see this failure on Linux when I did the release; seems like it would be deterministic and dune-release forces the tests to be run

Issue #117, opened 2024-06-24
file descriptor returned by Uring.accept is unusable


koonwen:
The Uring `accept` API returns the file descriptor for the accepted socket in `result:int` as part of 
```ocaml
 type 'a completion_option =
  | None
  | Some of { result: int; data: 'a }
```
However, since the type of `Unix.file_descr` is abstract, it's unusable with the Unix module API's. My work around is to just use `Obj.magic` to convert it to the `file_descr` type, but is this the correct usage pattern?

talex5:
Yes (but check it isn't negative first).

(see also https://discuss.ocaml.org/t/detect-use-after-close-of-unix-file-descriptors/13766/8)

talex5:
Fixed in #131.

PR #129, opened 2025-10-02
Stop recommending wizardry when dealing with fds

dra27:
Provides a solution for #117, but at a cost of an extra C call.


avsm:
Holding this until the interesting design discussion in #130 is resolved.

talex5:
I've modified this approach slightly in #131 - you can now only convert uring results to FDs, not any int.

PR #130, opened 2025-10-02
Type-safe wait functions

dra27:
This is a more invasive alternative to #129 (which also "fixes" #117). I've pontificated about it [on my blog](https://www.dra27.uk/blog/platform/2025/10/01/int-file-descriptors-considered-harmful.html).

With this idea, even in a thin wrapping around io\_uring, an `int` in OCaml is only used to represent an actual (positive-or-zero, as it happens) `int`, not an encoding of something else. For the fast-path (i.e. where an error is not being returned) there is _no_ difference in allocations, and unlike #129, no cost to retrieving the `Unix.file_descr` value from syscalls which return descriptors.

I didn't try plumbing it all the way back to Eio, but if moving the API in this general direction is an agreeable idea, I'd be quite happy to (it might finally tempt me to get my dirty Windows paws on eio backends as well...).

- [ ] Differentiate unit and int?
- [ ] Store real IDs in `Heap`, and simplify `'a job`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,13 +59,16 @@ We can now ask Linux to suspend the process until a result (Completion Queue Ent
 let rec wait_with_retry uring =
   match Uring.wait uring with
   | None -> wait_with_retry uring        (* Interrupted *)
-  | Some { result; data } -> result, data;;
+  | Some { result; kind = Uring.Int; data } -> None, (result : int), data
+  | Some { result; kind = Uring.FD; data } -> Some (result : Unix.file_descr), 0, data
+  | Some { result; kind = Uring.Error; data } -> failwith (Unix.error_message result)
 ```
 
 <!-- $MDX non-deterministic=output -->
 ```ocaml
-# let result, data = wait_with_retry uring;;
-val result : int = 8
+# let fd, result, data = wait_with_retry uring;;
+val fd : Unix.file_descr option = Some <abstr>
+val result : int = 0
 val data : _[> `Open_log ] = `Open_log
 ```
 
@@ -75,16 +78,10 @@ The `data` field is the data we passed in when submitting the request, allowing 
 The `result` field is the return code,
 with the same meaning as the return code from the corresponding system call (`openat2` in this case).
 
-```ocaml
-# let fd =
-    if result < 0 then failwith ("Error: " ^ string_of_int result);
-    (Obj.magic result : Unix.file_descr);;
-val fd : Unix.file_descr = <abstr>
-```
-
 We can now submit further requests. e.g.
 
 ```ocaml
+let fd = Option.get fd
 let rec write_all fd = function
   | [] -> ()
   | bufs ->
@@ -97,7 +94,7 @@ let rec write_all fd = function
       |> Option.get               (* We know we have enough space here *)
     in
     assert (Uring.submit uring = 1);
-    let result, data = wait_with_retry uring in
+    let _, result, data = wait_with_retry uring in
     assert (data = `Write_all);  (* There aren't any other requests pending *)
     assert (result > 0);         (* Check for error return *)
     let bufs = Cstruct.shiftv bufs result in
@@ -124,8 +121,9 @@ Some <abstr>
 - : int = 1
 
 # wait_with_retry uring;;
-- : int * ([> `Close_log | `Open_log | `Write_all ] as '_weak3) =
-(0, `Close_log)
+- : Unix.file_descr option * int *
+    ([> `Close_log | `Open_log | `Write_all ] as '_weak3)
+= (None, 0, `Close_log)
 ```
 
 The file has now been written:
--- a/README.md
+++ b/README.md
@@ -59,16 +59,16 @@ We can now ask Linux to suspend the process until a result (Completion Queue Ent
 let rec wait_with_retry uring =
   match Uring.wait uring with
   | None -> wait_with_retry uring        (* Interrupted *)
-  | Some { result; kind = Uring.Int; data } -> None, (result : int), data
-  | Some { result; kind = Uring.FD; data } -> Some (result : Unix.file_descr), 0, data
-  | Some { result; kind = Uring.Error; data } -> failwith (Unix.error_message result)
+  | Unit { result = (); data } -> Either.Left 0, data
+  | Int { result; data } -> Either.Left result, data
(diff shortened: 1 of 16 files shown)


avsm:
I'm a big plus for this! Having ocaml-uring return well-typed Unix.file_descrs is a perfect place to put the abstraction boundary, I feel. It is still useful to be able to convert this to a raw integer for other uses though (like passing it to programs that want an fd number)

dra27:
Does that fit in with [this related thought](https://www.dra27.uk/blog/platform/2025/09/30/file-descriptors-are-not-integers.html)? Would it be the case that you _must_ pass the fd to a program with its current number, or could you pick one (and have `Unix.create_process` or whatever handle the `dup` machinery)?

avsm:
File descriptors are very much integers on *nix' they're just not on Windows! It's not exactly common, but programs do occasionally just need the fd integer for their own purposes, especially fd passing.

dra27:
When’s the situation where you need to pass a specific fd _as that fd_, though? That was the point I was coming around to - if the file descriptor were abstract in C as well, when precisely would that hurt. Process invocation is one, but solvable with a map. Are there concrete cases where it actually can’t be done?

dra27:
The distinction I was trying to draw is that obviously there are times in C where you need to get at the C representation of the file descriptor - but are there situations other than process spawning where you _have_ to be able to get at it in OCaml?

dra27:
> File descriptors are very much integers on *nix' they're just not on Windows!

Because the hair-splitting is relevant here: _OCaml_ file descriptors are not integers on Windows - the _C_ ones very much are, the _OS_ ones are pointers (where on Unix, the OS and C descriptors share the same thing)

talex5:
> Does that fit in with [this related thought](https://www.dra27.uk/blog/platform/2025/09/30/file-descriptors-are-not-integers.html)?
> The fix is easy: pass a (int * file_descr) list. We preserve the abstraction, and specify the mapping.

Note: that is very similar to what Eio already does:

https://github.com/ocaml-multicore/eio/blob/7695d22387af7dc98214d66ad003dfad7b2a38f8/lib_eio/unix/process.mli#L27

avsm:
@dra27 wrote:
> Because the hair-splitting is relevant here: OCaml file descriptors are not integers on Windows - the C ones very much are, the OS ones are pointers (where on Unix, the OS and C descriptors share the same thing)

Ah I understand your point now. Right now we have:

- linux: (int * int:fd )
- windows: (int * struct:fd)

Where fd (Unix.file_descr) can be either an int or a struct.

However, this io_uring library will _only ever_ be run on Linux, and so imposing an overhead of allocating a mapping tuple on Linux-only code seems unnecessary -- the two contents of the tuple will be identical in their representation! We could instead expose a `type unix_fd = private int` from this library, use that directly in Eio_linux (which will never see a Windows fd) and then turn that into a Unix.file_descr in the portable Eio layers.

dra27:
Yes, indeed, a `private int` would be a good alternative - although where in Eio would one of these instances be? I've only had a cursory look at low_level.ml and sched.ml in lib_eio_linux, so I'm probably missing it, but I can see the parts where it converts an `int` from the result back to a `file_descr` (via Eio's own fds) but I hadn't yet found the reverse (but I should probably just try pushing this through to see!)?

I've only just started looking at it, but given the remarkable similarities between liburing and `ioringapi.h`, I'm not sure this library necessarily has to stay Linux only! 🧐

avsm:
I'm not sure where to plumb it into Eio (which might overuse Unix.file_descr in the Linux backend because it's convenient). But that shouldn't stop us from doing it right in this library, and putting the conversion into Eio (which could eventually have a bigger refactor to make it more amenable to the backend-specific fd tracking)

avsm:
@dra27 wrote:
> I've only just started looking at it, but given the remarkable similarities between liburing and ioringapi.h, I'm not sure this library necessarily has to stay Linux only! 

I think they are very similar on the surface (https://windows-internals.com/ioring-vs-io_uring-a-comparison-of-windows-and-linux-implementations/ is helpful), but there are lots of differences as the more advanced pieces of kernel plumbing get involved (like fixed buffer handling). I'd probably create a separate Windows only ioring library, and then do the integration in Eio in the first instance.

talex5:
So to summarise this PR:

- When submitting a job, we encode the expected return type in 4 bits of the (C code's) job ID.
- When getting a result, the C stubs use this to create an OCaml block with a tag appropriate to the type.
- uring.ml then converts this to a similar typed variant to hand to the user (but now with the user data).

This means that users get suitable return types (e.g. `Unix.file_descr` or `error`) rather than `int`, which is clearly an improvement.

The cost is that jobs are no longer ints and now require allocations (which `Heap` was trying to avoid IIRC) because we need to store both the OCaml ID and the full C ID (needed for cancellation). It might be better to add the type encoding on the OCaml side to avoid this. I guess this is what you mean by "Store real IDs in Heap, and simplify 'a job?" - sounds good to me.

I could have a go at this if you want.

(note: there are also some `AT_FDCWD` changes from #128 included in this PR)

talex5:
From a quick go at trying this with Eio:

- Some error codes aren't known to OCaml. e.g. `ETIME` would become `EUNKNOWNERR 62`. But this is fragile; if OCaml ever adds the error codes, the match will silently stop working.
- Currently we resume fibers by passing an `int` with the result. Now we have to allocate a `result` (or raise, but then we need to know the name of the operation too, and the stack trace shows the error coming from the scheduler).
- The matching gets more complicated, because we have to remember to match both the success and error cases for every type.

So, not clear this is an improvement for Eio (though with the current API you can forget to check for errors, but that could be fixed with an `nat_or_error` type).

talex5:
Thanks for making these experiments!

I've gone with the simpler version in #131. It would be possible to combine the approaches, so that you can't get an FD from a non-FD syscall, but that isn't a mistake anyone's likely to make (it would never work) so probably not worth it, and it's easy enough to get invalid FDs in other ways anyway (e.g. with `Unix.close`).

I'll close this draft for now, but feel free to reopen if you still want to finish it. Based on trying this API with Eio, I think the main useful thing would be to avoid the separate `Error` return, since it's easy to forget to handle that (while the separate int-or-error vs fd-or-error is OK). It might also be easier to make the change entirely on the OCaml side (encoding the type into the heap ptr) and leave the C stubs as they are.

PR #131, opened 2026-05-12
Add Uring.Res module to handle result values

talex5:
This is a modified version of @dra27's #129. Instead of returning `int`, uring operations now return values of type `Uring.Res.t = private int`. The `Res` module provides functions for converting to ints or FDs, as results or raising exceptions.

This means:
- You can't forget to handle errors, and handling them is now easier.
- No use of `Obj.magic` in the examples or tests.
- There's a pretty-printer for results, so the tests and top-level no longer show negative numbers for errors.

Unlike #129, there's no overhead of an extra C call and we don't provide a way of turning arbitrary ints into FDs (only the results of uring operations). Though given the design of `Unix` I still don't see why it shouldn't provide a way to do that anyway.

Unlike #130, existing code needs only minor changes (in particular, you can just cast the result to `int` and carry on as before if it's too annoying). It is now possible to get an FD from a result that wasn't supposed to be one (e.g. `read`), but that doesn't seem like a mistake anyone's likely to make.

Annoyingly, OCaml's Unix module doesn't expose a way to turn error codes into enum names, even though it does so itself as part of its exception printer, so I used `Unix.error_message` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,9 +76,7 @@ The `result` field is the return code,
 with the same meaning as the return code from the corresponding system call (`openat2` in this case).
 
 ```ocaml
-# let fd =
-    if result < 0 then failwith ("Error: " ^ string_of_int result);
-    (Obj.magic result : Unix.file_descr);;
+# let fd = Uring.Res.fd_exn result "openat2" "test.log";;
 val fd : Unix.file_descr = <abstr>
 ```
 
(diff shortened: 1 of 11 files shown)


talex5:
> Annoyingly, OCaml's Unix module doesn't expose a way to turn error codes into enum names, even though it does so itself as part of its exception printer, so I used Unix.error_message instead.

And this is a problem, because musl produces different strings. I've added a custom printer to fix that.

PR #142, opened 2026-05-29
Statx attribute supported check was inverted

avsm:
I dont think we use this in Eio, so wasnt noticed. The new tests (d5af414f56d69b7ff72b02a37291555c75adb509) fail if the fix (77327a36de61c5e3ad0e174de68bdbcc6436537d) is reverted.


tests/main.md
```diff
@@ -287,6 +298,80 @@ val fd : unit = ()
...
+  assert (not (Flags.mem Flags.symlink_nofollow Flags.empty));;
+- : unit = ()
+
+# let combo = Flags.(symlink_nofollow + no_automount) in
+  assert (Flags.mem Flags.symlink_nofollow combo);
+  assert (Flags.mem Flags.no_automount combo);
+  assert (not (Flags.mem Flags.empty_path combo));;
+- : unit = ()
+
+# assert ((Flags.(symlink_nofollow + symlink_nofollow) :> int) = (Flags.symlink_nofollow :> int));
```
talex5: I assume the cast here is needed for OxCaml? Maybe we should have a `Flags.(=)`?

tests/main.md
```diff
@@ -287,6 +298,80 @@ val fd : unit = ()
...
+`Mask.basic_stats` is the union of the basic components, but excludes `btime` (birth time):
+
+```ocaml
+# let components = Mask.[ type'; mode; nlink; uid; gid; atime; mtime; ctime; ino; size; blocks ] in
+  List.iter (fun c -> assert (Mask.mem c Mask.basic_stats)) components;
+  assert (not (Mask.mem Mask.btime Mask.basic_stats));;
+- : unit = ()
+
+# let m = Int64.of_int (Mask.(size + type') :> int) in
+  assert (Mask.check m Mask.size);
```
talex5: This `check` API is a bit puzzling. Why is it taking masks and flag sets as int64 rather than as the abstract type?

I think I'd expect:
```ocaml
    val check : ?mask:t -> t -> t -> bool
```

Also, it's a shame that the arguments are the opposite way around from `Attr.mem`. I think it could be implemented as something like:

```ocaml
    let check ?mask set t =
      mask |> Option.iter (fun mask -> if not (mem t mask) then invalid_arg "Attribute not supported");
      mem t set
```

talex5 (commented):
The fix looks right, but it's making me think that the `Attr` API needs revising.

tests/main.md
talex5: @patricoferris : does changing this sound sensible to you?

avsm: I've pushed this suggestion in 46fb3fdf5429e554c4ec325adb7900905d468f89 and also removed the int64<->int conversions, since (afaict) statx attributes leave plenty of room for 63 bits:

```
#define STATX_ATTR_COMPRESSED           0x00000004 /* [I] File is compressed by the fs */
#define STATX_ATTR_IMMUTABLE            0x00000010 /* [I] File is marked immutable */
#define STATX_ATTR_APPEND               0x00000020 /* [I] File is append-only */
#define STATX_ATTR_NODUMP               0x00000040 /* [I] File is not to be dumped */
#define STATX_ATTR_ENCRYPTED            0x00000800 /* [I] File requires key to decrypt in fs */
#define STATX_ATTR_AUTOMOUNT            0x00001000 /* Dir: Automount trigger */
#define STATX_ATTR_MOUNT_ROOT           0x00002000 /* Root of a mount */
#define STATX_ATTR_VERITY               0x00100000 /* [I] Verity protected file */
#define STATX_ATTR_DAX                  0x00200000 /* File is currently in DAX state */
#define STATX_ATTR_WRITE_ATOMIC         0x00400000 /* File supports atomic write operations */
```

tests/main.md
avsm: This was because the type was a private int, but I just made it abstract and added of/to_int and added a Flag.(=)

lib/uring/uring.ml
```diff
@@ -36,10 +38,14 @@ module Flags = struct
...
   let of_int x = x
 
+  let to_int x = x
+
   let ( + ) = ( lor )
 
   let mem a b =
     (a land b) = a
+
+  let ( = ) (a : t) (b : t) = Int.equal a b
```
talex5, line 48:
```suggestion
  let ( = ) = Int.equal
```

lib/uring/uring.mli
```diff
@@ -604,19 +611,16 @@ module Statx : sig
         @see <https://www.kernel.org/doc/Documentation/filesystems/dax.txt> Direct Access for Files
         @since Linux 5.8 *)
 
-    val check : ?mask:Int64.t -> Int64.t -> t -> bool
-    (** [check ?mask attr t] will check if [t] is set in [attr].
-
-        If [mask] is not [None] then it will first check the mask to see
-        if the file attribute is supported and if not raise [Invalid_argument]. *)
+    val check : ?mask:t -> t -> t -> bool
```
talex5: How about renaming this to `mem_checked`, switching the argument order to match, and making `mask` non-optional?

I guess every existing user can just use `mem`, since `mask` never worked.

(what is this for anyway? why would someone want an `Invalid_argument` exception if an attribute isn't supported? isn't it better just to report that the attribute isn't set in that case?)

lib/uring/uring.mli
```diff
@@ -25,18 +25,25 @@ module Region = Region
 
 (** Type of flags that can be combined. *)
 module type FLAGS = sig
-  type t = private int
+  type t
```
talex5, line 28: Is this to allow for non-int types in future?

lib/uring/uring.mli
avsm: The kernel API here is a little odd:

    It should be noted that the kernel may return fields that weren't
    requested and may fail to return fields that were requested,
    depending on what the backing filesystem supports.  (Fields that
    are given values despite being unrequested can just be ignored.)
    In either case, stx_mask will not be equal mask.

So the caller needs to be sure that the mask is set as an invariant before using those values. This might be expressed better as a GADT over what fields the user actually wants, and hide all this.

talex5: There seem to be two masks here:
- `stx_mask` is "Mask of bits indicating filled fields"
- `stx_attributes_mask` is "Mask to show what's supported in `stx_attributes`"

Checking `stx_mask` is important, since otherwise the field may give wrong information.

`stx_attributes` are just flags and it looks less important there, although I see the man-page does say "Note that any attribute that is not indicated as supported by `stx_attributes_mask` has no usable value here", so I guess we should check it.

But it seems odd to me that the kernel might say that e.g. `STATX_ATTR_COMPRESSED` isn't supported, while at the same time having it turned on.

lib/uring/uring.ml
avsm: OxCaml has stricter eta expansion requirements. It would be easier to leave this expanded and avoid breaking there with the future interaction with `local`

lib/uring/uring.mli
avsm: So I guess the only semantic that makes sense is that to be usable, it must be active in _both_ `stx_attributes` and `stx_mask` for a given statx reply.

lib/uring/uring.mli
avsm: Mainly to allow wiggle room for oxcaml variants (like small ints). It serves no useful purpose in this OCaml one, but if it's not a big deal otherwise it seems better to just make it abstract.

lib/uring/uring.mli
avsm: done in 89cf92f4dc5044b60f63b22e7f5d86d06ba91054. I considered a polymorphic variant to distinguish unset/unsupported, but looking at the kernel source code I can't spot how this would ever happen so it doesnt seem worth checking for.

lib/uring/uring.ml
talex5: I'd expect possible problems with partial application, but I'm surprised if simply making an alias is a problem.

talex5:
Squashed and rebased. I also removed the warning about running the tests on "a compressed or immutable filesystem", as I don't think that applies now (it's just testing pure functions, not actually using the FS).

PR #143, opened 2026-05-29
regen primops and make lintcstubs-arity a test depopt

avsm:
I missed this depopt, which in turn didn't regen the primitives.h. I *think* depopts supports with-test, so this should catch future missing regenerations in CI

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -22,6 +22,5 @@
   (cmdliner (and :with-test (>= 1.1.0)))
   (fmt (>= 0.8.10))
   (optint (>= 0.1.0))
-  (mdx (and (>= 2.1.0) :with-test)))
- (depopts
+  (mdx (and (>= 2.1.0) :with-test))
   (lintcstubs-arity :with-test)))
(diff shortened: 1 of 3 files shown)


talex5:
Does it even need to be optional? I'm trying to remember why it wasn't included by default. Maybe it didn't support all compiler versions back then?

avsm:
It would make it easier to test Uring on unreleased compilers where the cmt parsing has changed. Marginally useful to keep it optional, I think.

talex5:
I don't see `lintcstubs-arity` being installed in the CI output.

talex5:
> It would make it easier to test Uring on unreleased compilers where the cmt parsing has changed. Marginally useful to keep it optional, I think.

It could still be optional in the dune file, but not in opam, which would make testing on an unreleased compiler fairly easy (just install the other test deps manually, or remove that line from opam).

avsm:
Confirmed its installing now in CI, so merging

```
-> retrieved lintcstubs-arity.0.5.2  (cached)
```

PR #144, opened 2026-05-29
Report errors from io_uring_submit

talex5:
Previously, it just returned a negative number on error.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -14,7 +14,7 @@
  (description "Bindings to the Linux io_uring kernel IO interfaces. See https://github.com/ocaml-multicore/eio for a higher-level API using this.")
  (depends
   (cstruct (>= 6.0.1))
-  (ocaml (>= 4.12.0))
+  (ocaml (>= 5.2.0))
   dune-configurator
   (lwt (and :with-test (>= 5.0.0)))
   (bechamel (and (>= 0.1.0) :with-test))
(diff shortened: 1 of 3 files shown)


avsm (approved):
Looks fine; I had some problems with odd EINTRs being returned in very early iterations of this back in 5.x kernels, but I don't think we care about those any more.

avsm:
This won't work on OCaml 4.x, but why don't we just make this library OCaml 5.x only now? I'm only practically using this in Eio which is obviously 5.x only.

```
File "lib/uring/dune", line 8, characters 9-20:
8 |   (names uring_stubs)
^^^^^^^^^^^
(cd _build/default/lib/uring && /usr/bin/gcc -O2 -fno-strict-aliasing -fwrapv -pthread -fPIC -D_FILE_OFFSET_BITS=64 -Wall -fdiagnostics-color=always -g -I /home/opam/.opam/4.14/lib/ocaml -I /home/opam/.opam/4.14/lib/cstruct -I /home/opam/.opam/4.14/lib/fmt -I /home/opam/.opam/4.14/lib/optint -I include -o uring_stubs.o -c uring_stubs.c)
uring_stubs.c: In function 'ocaml_uring_submit':
uring_stubs.c:937:5: error: implicit declaration of function 'caml_unix_error'; did you mean 'unix_error'? [-Wimplicit-function-declaration]
937 |     caml_unix_error(-num, "io_uring_submit", Nothing);
|     ^~~~~~~~~~~~~~~
|     unix_error

```

talex5:
> Looks fine; I had some problems with odd EINTRs being returned in very early iterations of this back in 5.x kernels, but I don't think we care about those any more.

Even if so, getting a negative number back from `submit` isn't useful.

> why don't we just make this library OCaml 5.x only now?

Good idea. Done!

PR #145, opened 2026-05-30
Add bounds check to Region.init

avsm:
Might as well raise early here than a subsequent access

### DIFF
--- a/lib/uring/region.ml
+++ b/lib/uring/region.ml
@@ -12,17 +12,10 @@ type chunk = t * int
 
 exception No_space
 
-let init ~block_size buf slots =
-  if block_size < 0 then
-    invalid_arg (Printf.sprintf "Region.init: negative block_size %d" block_size);
-  if slots < 0 then
-    invalid_arg (Printf.sprintf "Region.init: negative slots %d" slots);
-  let needed = block_size * slots in
-  let available = Bigarray.Array1.dim buf in
-  if needed > available then
-    invalid_arg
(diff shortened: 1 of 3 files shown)


lib/uring/region.ml
```diff
@@ -13,6 +13,16 @@ type chunk = t * int
 exception No_space
 
 let init ~block_size buf slots =
+  if block_size < 0 then
+    invalid_arg (Printf.sprintf "Region.init: negative block_size %d" block_size);
+  if slots < 0 then
+    invalid_arg (Printf.sprintf "Region.init: negative slots %d" slots);
+  let needed = block_size * slots in
```
avsm: I didn't do an integer overflow check here; is it worth it?

talex5: Can't we just calculate the number of slots from the buffer size and slot size? That would avoid the multiplication completely.

avsm: I'd considered that, but it does change the API; for example you can choose to underfill a buffer right now. However, I'm not sure what use underfilling the passed in buffer would be (page alignment? Our slots are a multiple anyway). And given the main consumer is eio, we can easily just bump the uring minimum version there with a changed API. Pushed in 907e8a49bc0513eef21caf2ca8f359e60a0c4923

talex5 (approved):

lib/uring/region.ml
talex5: We should probably just move or copy `Uring.Region` into Eio anyway.

PR #147, opened 2026-05-30
Add shutdown, socket, renameat and symlinkat

avsm:
Each commit is separate with associated tests. Also fixed the symlink-follow test while I was there as it works with the latest liburing.

### DIFF
--- a/lib/uring/uring_stubs.c
+++ b/lib/uring/uring_stubs.c
@@ -1161,3 +1161,35 @@ ocaml_uring_submit_linkat_byte(value* values, int argc) {
     values[6]
   );
 }
+
+value /* noalloc */
+ocaml_uring_submit_renameat_native(value v_uring, value v_id,
+    value v_old_dir, value v_old_path,
+    value v_new_dir, value v_new_path,
+    value v_flags) {
+  struct io_uring *ring = Ring_val(v_uring);
+  char *old_path = Sketch_ptr_val(v_old_path);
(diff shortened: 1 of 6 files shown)


lib/uring/uring_stubs.c
```diff
@@ -865,6 +865,41 @@ ocaml_uring_submit_accept(value v_uring, value v_id, value v_fd, value v_sockadd
...
+};
+
+value /* noalloc */
+ocaml_uring_submit_socket(value v_uring, value v_id, value v_domain, value v_type, value v_protocol) {
+  struct io_uring *ring = Ring_val(v_uring);
+  struct io_uring_sqe *sqe = io_uring_get_sqe(ring);
+  if (!sqe) return (Val_false);
+  io_uring_prep_socket(sqe,
+                       socket_domain_table[Int_val(v_domain)],
+                       socket_type_table[Int_val(v_type)],
```
talex5: Might be worth checking these are in-range. I can imagine the compiler might add more at some point.

talex5 (approved):
I rebased this and also hard-coded the rename flags (for musl).

lib/uring/uring_stubs.c
avsm: Yes good point but I'd like to do this once in discover.ml rather than every invocation. Coming up...

avsm: the easier way was just to reuse the runtime table as an external! f20e359b9b98629c01ea6a7c15587d09d1801fe0

PR #148, opened 2026-06-02
caml_uring_wait_cqe should have the same null cqe check that the timeout variant has

avsm:


avsm:
Actually, thinking about this again, I'm not 100% sure that the non-timeout version would ever return a NULL cqe...

talex5 (commented):
Yes, I think this shouldn't be needed. `_io_uring_get_cqe` returns:
- Reporting an error if `__io_uring_peek_cqe` or `__sys_io_uring_enter2`  fails.
- If we didn't ask to wait for anything (but we do here).
- If `need_enter` is false (but it's always true if `nr_available < 1`, and otherwise we have a CQE).
- On timeout (but we don't have one).
- If `cqe` isn't NULL.

But the logic is a bit complicated, and it wouldn't hurt to include this check for safety.

avsm:
What do you think about making this a C assert instead? We ought to know if this ever triggers, but it should never trigger.

talex5:
> What do you think about making this a C assert instead?

Sounds good to me.

lib/uring/uring_stubs.c
```diff
@@ -1007,6 +1007,8 @@ value ocaml_uring_wait_cqe(value v_uring)
       unix_error(-res, "io_uring_wait_cqe", Nothing);
     }
   } else {
+    if (!cqe)
+      CAMLreturn(Val_cqe_none);
```
avsm:
```suggestion
    assert (cqe != NULL);
```

lib/uring/uring_stubs.c
avsm, line 41:
```suggestion
#include <sys/sysmacros.h>
#include <assert.h>
```

PR #149, opened 2026-06-02
add fallocate and ftruncate bindings

avsm:


talex5 (commented):
Looks good, but CI is failing because `FALLOC_FL_WRITE_ZEROES` is very new and most distros don't have it.

avsm:
I think our CI  boxes are on an old kernel and so it's getting EINVAL despite the userspace (like ubuntu 26.04) supporting it. will take a look

talex5:
OK, but uring needs to be installable on older systems too. Probably we should just commit the generated module into the source tree.

avsm:
Agreed; pushed in d820ff0ee7f9d6339abec5c6171b247bf12862e0 and tests adjusted to skip on older kernels in 21965ec3b30c166064de7bc6ee0efe81b8900d3f

PR #151, opened 2026-06-05
allow setting socket flags on socket creation

avsm:
Allow setting CLOEXEC to the Uring socket creation, so that https://github.com/ocaml-multicore/eio/pull/843 can be done directly from uring as well


tests/main.md
```diff
@@ -970,6 +970,54 @@ val fd : unit = ()
...
+
+```ocaml
+# let cloexec_set fd =
+    let ic = open_in (Printf.sprintf "/proc/self/fdinfo/%d" fd) in
+    Fun.protect ~finally:(fun () -> close_in ic) @@ fun () ->
+    let rec loop () =
+      match input_line ic with
+      | line ->
+        (match String.split_on_char ':' line with
+         | ["flags"; v] -> (int_of_string ("0o" ^ String.trim v)) land 0o2000000 <> 0
```
avsm, line 987: Suggestions on better ways to do this are welcome! This dance is because we don't have fcntl yet

lib/uring/uring.mli
```diff
@@ -289,6 +289,17 @@ module Open_flags : sig
       on other file types is unspecified. *)
 end
 
+(** Flags that can be passed to {!socket} when creating a socket. *)
+module Sock_flags : sig
+  include FLAGS
+
+  val cloexec : t
+  (** [cloexec] sets the close-on-exec flag on the new socket. *)
```
talex5, line 297: It might be safer to do this always, rather than offering it as a flag. I think opening things non-cloexec is only useful for backwards compatibility (which we don't need to care about), and you can always change it afterwards if you really want to.

avsm: I think the uring bindings should follow kernel defaults where possible; why not just always set this from Eio_linux? The user can't issue direct uring requests anyway without going through Eio.

talex5: That's possible too. I just dislike it for close-on-exec because it's so easy to get it wrong and not notice, and I'm sure the kernel only uses this API to preserve backwards compatibility with old C code.

avsm: It is a bad default in the modern world, I agree. I'll change this so that the flags arg is an optional labeled argument, with the default setting the CLOEXEC flag. This way a caller can just do `~flags:[]` to go to the default behaviour and we can avoid having a double-negative flag (`NO_CLOEXEC`).

avsm: done in 1720359dcd725b25f08b164eb300ae0b3a93d9a8

PR #152, opened 2026-06-05
add an errno module with Linux specific errnos

avsm:
Some weird errnos in here!

### DIFF
--- a/lib/uring/uring.ml
+++ b/lib/uring/uring.ml
@@ -421,7 +421,7 @@ module Errno = struct
     | `EUNKNOWN of int
   ]
 
-  (* The Linux asm-generic errno table deroived from 
+  (* The Linux asm-generic errno table derived from
      [https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/errno.h]. *)
   let table = [
     1, `EPERM, "EPERM";       2, `ENOENT, "ENOENT";   3, `ESRCH, "ESRCH";
@@ -513,10 +513,6 @@ module Errno = struct
 
   let pp f e = Fmt.string f (to_string e)
 
-  let is_retry : t -> bool = function
-    | `EINTR | `EAGAIN | `EWOULDBLOCK | `ETIME -> true
-    | _ -> false
-
   let to_unix e : Unix.error = error_of_errno (to_int e)
 
   let of_unix : Unix.error -> t = function
@@ -549,6 +545,18 @@ module Errno = struct
     | EUNKNOWNERR n -> of_int n
 end
 
+exception Linux_error of Errno.t * string * string
+
+let to_unix_error = function
+  | Linux_error (e, fn, arg) -> Unix.Unix_error (Errno.to_unix e, fn, arg)
+  | exn -> exn
+
+let () =
+  Printexc.register_printer (function
+    | Linux_error (e, fn, arg) ->
+      Some (Printf.sprintf "Uring.Linux_error(%s, %S, %S)" (Errno.to_string e) fn arg)
+    | _ -> None)
+
 module Res = struct
   type t = int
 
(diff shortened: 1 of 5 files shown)


lib/uring/uring.ml
```diff
@@ -380,9 +380,173 @@ end
...
+    | `ENOTNAM | `ENAVAIL | `EISNAM | `EREMOTEIO | `EDQUOT | `ENOMEDIUM
+    | `EMEDIUMTYPE | `ECANCELED | `ENOKEY | `EKEYEXPIRED | `EKEYREVOKED
+    | `EKEYREJECTED | `EOWNERDEAD | `ENOTRECOVERABLE | `ERFKILL | `EHWPOISON
+    (* Alias names that share a code with one of the above *)
+    | `EWOULDBLOCK | `EDEADLOCK | `EFSBADCRC | `EFSCORRUPTED
+    (* Any code we don't have a name for *)
+    | `EUNKNOWN of int
+  ]
+
+  (* The Linux asm-generic errno table deroived from
```
talex5:
```suggestion
  (* The Linux asm-generic errno table derived from 
```

lib/uring/uring.mli
```diff
@@ -101,12 +101,71 @@ module Setup_flags : sig
...
+  val of_unix : Unix.error -> t
+  (** [of_unix e] lifts a {!Unix.error} into a [t] value. *)
+
+  val to_unix : t -> Unix.error
+  (** [to_unix t] is the corresponding {!Unix.error}. Linux-specific codes
+      with no {!Unix.error} name become [EUNKNOWNERR n]. *)
+
+  val is_retry : t -> bool
+  (** [is_retry t] is [true] for codes indicating the operation should be
+      be reissued. *)
```
talex5: I'm not sure this is actually useful. Eio handles these to retry fixed-buffer operations (only), but I don't know why the kernel would return them. Possibly it was just a bug in old versions of uring?

(there is https://github.com/ocaml-multicore/eio/issues/788 but that only needs EINTR)

I suggest removing it.

lib/uring/uring.mli
```diff
@@ -101,12 +101,71 @@ module Setup_flags : sig
...
 module Res : sig
   (** The result of a uring operation (typically the same as the return value of the corresponding syscall). *)
 
   type t = private int
   (** The return value. If negative, it is the negative errno value giving the error. *)
 
+  val errno : t -> Errno.t option
+  (** [errno t] is [Some e] if [t] is negative (an error), or [None] otherwise. *)
+
   val int_result : t -> (int, Unix.error) result
```
talex5:
```suggestion
  val int_result : t -> (int, Errno.t) result
```
and the same for FDs.

lib/uring/uring.mli
talex5: (I'll try removing it from Eio in https://github.com/ocaml-multicore/eio/pull/851)

lib/uring/uring.ml
avsm: fixed in adf5ca1b9cbc7532ecc8936adb91a8e78970c20f

lib/uring/uring.mli
avsm: removed in adf5ca1b9cbc7532ecc8936adb91a8e78970c20f

lib/uring/uring.mli
avsm: What do you think of 3b8fa7e753a79a43f31e91374eadc8ac5260d09d now? The get_cqe is a tiny bit slower due to reraising errors.

We could raise the errors from C easily if we convert away from a polymorphic variant here, but it's not that much slower to install an exception handler (but these things do add up I guess).

avsm:
Hm @talex5, this PR isn't actually that much use in Eio, since the Low_level part of Eio might fall back to using Unix module functions, and so we still have to deal with Unix_error. And if you don't need to the retry functionality anyway, there's much else utility left in this PR.

lib/uring/uring.mli
talex5: I think I prefer the exception-raising versions with `Unix_error`, as they were before. That is the normal exception that people will expect. I'd guess that people who want to handle specific errors will use the `_result` functions anyway (and if you do want to handle unusual codes in exceptions, you can still convert the error code manually if necessary).

avsm: Yes I tried to switch to a richer exception, but there's no point as Unix_error is always necessary. I'll switch to stashing the Linux specific errors in EUNKNOWNERR

talex5:
> this PR isn't actually that much use in Eio,

Hmm, true. If we do find some useful ones, it might be worth adding them upstream to OCaml instead anyway. I'll close this for now then.

avsm:
Yeah I'll come back to it if useful in the future.

PR #153, opened 2026-06-06
switch the buffer type to bytes instead of bigarray/cstruct

avsm:
This is a surprisingly small diff, since most of what we need to do is just make sure the backing `bytes` is >2KB.


talex5:
Does benchmarking show an improvement?

avsm:
"Better", but I'm working on quantifying this. Stay tuned.

avsm:
I ran some benchmarks against both branches, and the microbenchmarks show:

- no regression in I/O throughput for the filesystem benchmarks, and an obvious improvement when the library uses bytes (and so avoids a Cstruct.to_bytes as most of my e.g. http libraries do)
- improves large-buffer allocation cost >2kb is better with bytes  as the bigarray finalizer and C proxy is not needed anymore. Also I'd forgotten that Cstruct zeros buffers whereas we use Bytes.create which leaves them uninitialised (probably a better default given we're about to overwrite it anyway?)
- small buffers are worse with `bytes` since we clamp allocations to 2KB to make them immovable. Callers to this library will need to be careful to coalesce small writes before sending them to the kernel, but this is good practise anyway.

talex5:
> `Bytes.create` which leaves them uninitialised (probably a better default given we're about to overwrite it anyway?)

It only gets overwritten if there's enough data to fill the buffer. Otherwise you're leaking private data. Could have a `create_unsafe` though.

I'm wondering if we should have an `Iovec.to_cstruct` that shares the underlying memory. Then the buffer could be accessed by whatever API you had to use without any copying.

avsm:
I'm thinking that if the caller library provides buffers that are 2kb+, then we avoid this issue as the caller can design how to provide memory that meets their own criteria.  All we need to do in uring is to verify that any buffer coming in is of a suitable size, which seems like a cheap check to do.

avsm:
@talex5 wrote:
> I'm wondering if we should have an Iovec.to_cstruct that shares the underlying memory. Then the buffer could be accessed by whatever API you had to use without any copying.

This took a few iterations, but I have a glorious hack in 4116ced2ef40a686b0db319595561a3d53ea95ab that does successfully alias non-moving bytes. It create a proxy managed bigarray and a finaliser that ties the bytes to the bigarray. This scheme has the advantage that there's no performance impact if bigarrays aren't used (which is our eventual end goal).

Still stress testing it but it doesn't seem to obviously crash...

PR #154, opened 2026-06-12
add per-operation RWF flags to read/write submissions

avsm:
While rooting around in the innards of liburing, I discovered there are preadv2 and pwritev2 _flags_ supported in Linux that could save us a bunch of ops in Eio_linux.

Specifically, we could use the `dsync` flag to avoid an fsync op following a write, and the nocache op in Eio.Flow.copy to avoid polluting the buffer cache with the results of a large copy operation. This exposes the flags at the uring layer to start us off. Using them in Eio might involve a bit of an 'op supported' dance as some of the fun ones are in fairly recent 6.x kernels.


patricoferris (approved):
These changes look good to me, I can imagine they would be useful.

talex5:
Seems reasonable, but needs rebasing now that #155 is merged.

PR #155, opened 2026-06-14
add bindings for vectored read/writev fixed buffers

avsm:
Previously, the C stubs were misnamed and not actually doing vectored calls. Oops!


talex5 (approved):
Looks good!

(I split the rename into a separate commit just to make the diff easier to read)

PR #157, opened 2026-06-29
update to liburing.2.15

avsm:
Nothing relevant to our use that I could find, but some bugfixes and performance improvements, and the BPF support looks interesting for future support.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,9 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read
+
 
 jobs:
   get_commit_list:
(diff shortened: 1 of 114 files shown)


talex5 (approved):
Looks good!

(I combined the merge and the build-fix commits into one so that the history is always buildable)

talex5:
Oops, merged this too soon. It fails to install with opam's sandbox:
```
#=== ERROR while compiling uring.dev ==========================================#
# mktemp: failed to create file via template ‘/tmp/test_bpf_XXXXXX’: Read-only file system
```

Looks like this is the problem:

https://github.com/ocaml-multicore/ocaml-uring/blob/84192fb1905718405e4e6e5fced6fec3a9ada3e9/vendor/liburing/configure#L499

PR #158, opened 2026-07-01
Don't use /tmp in configure test

talex5:
opam's sandbox blocks access to this.

See https://github.com/ocaml-multicore/ocaml-uring/pull/157#issuecomment-4859602236


avsm:
looks good, may need to feed this upstream as well.

